PR #1, opened 2023-10-03
ARCH1: Add cinder backups

Akrog:
This patch adds the cinder backup using Ceph configuration to stage 6.

The documentation gdoc has new suggestions to create the "backups" pool and provide access to it.


abays (approved):
/lgtm

PR #7, opened 2023-10-06
Improved explanations of stage steps

abays:
Adds commands for each stage to demonstrate what must be done in individual steps

### DIFF
--- a/validated_arch_1/stage1/README.md
+++ b/validated_arch_1/stage1/README.md
@@ -4,9 +4,50 @@ Install the dependencies for Red Hat OpenStack Services on OpenShift operators
 
 ## Notes
 
-Requires an OpenShift 4.10+ cluster with 3 master/worker combo-nodes, with Metal3 and a provisioning network available
+Requires an OpenShift 4.12+ cluster with 3 master/worker combo-nodes, with Metal3 and a provisioning network available
+
+## Steps
 
 1. Create Namespaces
+```bash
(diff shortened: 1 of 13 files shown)


validated_arch_1/stage1/README.md
```diff
@@ -4,9 +4,50 @@ Install the dependencies for Red Hat OpenStack Services on OpenShift operators
...
+Requires an OpenShift 4.12+ cluster with 3 master/worker combo-nodes, with Metal3 and a provisioning network available
+
+## Steps
 
 1. Create Namespaces
+```bash
+oc apply -f cert_manager_namespace.yaml -f metallb_namespace.yaml -f nmstate_namespace.yaml
+```
 2. Create OperatorGroups
+```bash
```
afazekas, line 16: There multiple things which can be applied in single step from the directories.

Many cases basically just apply the a stage (or stages) in a loop (until dependent CRD definition arrives)  seams to be able to work.

abays: Yes, that is true.  The intent in showing things in separate steps was to clearly enumerate them for the user's understanding, but perhaps that is not necessary.

fultonj (approved):
/lgtm

validated_arch_1/stage1/README.md
abays: Let's decide one way or another in a potential follow-up, if necessary

PR #8, opened 2023-10-06
Add comment about MTU 9000 setting

fultonj:
Red Hat Ceph documentation recommends MTU 9000 (see the end of section 2.4 of [1]). Since enp7s0 is hosting Ceph traffic this setting of 9000 is correct.

Add a comment next to the setting stating that in our case there's one NIC, enp7s0 so we used that MTU. But if an environment has >1 NIC and the storage traffic is only on a particular NIC, then we could use a different MTU for other NICs.

[1] https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/6/html/hardware_guide/general-principles-for-selecting-hardware#network-considerations-for-red-hat-ceph-storage_hw

### DIFF
--- a/validated_arch_1/stage3/ocp_node_0_nncp.yaml
+++ b/validated_arch_1/stage3/ocp_node_0_nncp.yaml
@@ -76,6 +76,9 @@ spec:
         dhcp: false
       ipv6:
         enabled: false
+      # MTU 9000 is used to access Ceph via the storage network
+      # If storage network is on its own NIC adjust accordingly
+      # Using >1 NIC is not currently represented in these CRs
       mtu: 9000
       name: enp7s0 # CHANGEME
       state: up


raukadah:
+2 looks good o/

abays:
Are we not prescribing "single NIC w/ VLANs" as part of the architecture?  I had placed that in the list here: https://github.com/openstack-k8s-operators/architecture/tree/main/validated_arch_1#openstack-k8s-validated-architecture-one because I thought there was some sort of tentative consensus on that particular point of the networking consideration.

fultonj:
> Are we not prescribing "single NIC w/ VLANs" as part of the architecture? I had placed that in the list here: https://github.com/openstack-k8s-operators/architecture/tree/main/validated_arch_1#openstack-k8s-validated-architecture-one because I thought there was some sort of tentative consensus on that particular point of the networking consideration.

You are correct that we state we are using Network isolation over a single NIC which is a reasonable setting for a testing environment. A production environment could use >1 NIC. I think the VA can still be useful to someone with >1 NIC and that we could point this out. If however, we want to be extremely prescriptive we can do that.

abays:
> > Are we not prescribing "single NIC w/ VLANs" as part of the architecture? I had placed that in the list here: https://github.com/openstack-k8s-operators/architecture/tree/main/validated_arch_1#openstack-k8s-validated-architecture-one because I thought there was some sort of tentative consensus on that particular point of the networking consideration.
> 
> You are correct that we state we are using Network isolation over a single NIC which is a reasonable setting for a testing environment. A production environment could use >1 NIC. I think the VA can still be useful to someone with >1 NIC and that we could point this out. If however, we want to be extremely prescriptive we can do that.

I'm fine with including the comment as long as the user understands that using more than one NIC is not currently represented in the CRs (so they'd have to do more customization).

fultonj:
> > > Are we not prescribing "single NIC w/ VLANs" as part of the architecture? I had placed that in the list here: https://github.com/openstack-k8s-operators/architecture/tree/main/validated_arch_1#openstack-k8s-validated-architecture-one because I thought there was some sort of tentative consensus on that particular point of the networking consideration.
> > 
> > 
> > You are correct that we state we are using Network isolation over a single NIC which is a reasonable setting for a testing environment. A production environment could use >1 NIC. I think the VA can still be useful to someone with >1 NIC and that we could point this out. If however, we want to be extremely prescriptive we can do that.
> 
> I'm fine with including the comment as long as the user understands that using more than one NIC is not currently represented in the CRs (so they'd have to do more customization).

OK, I've added language to that effect in the latest patchset

abays (approved):
/lgtm

PR #10, opened 2023-10-09
Remove repo-setup from validated architecture 1

fultonj:
The repo-setup service is useful for installing upstream packages. Since validated architectures are supposed to for downstream users we should not include it. Add edpm_bootstrap_command example with CHANGEME in its place as described in dev preview documentation.

https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/18.0-dev-preview/html/deploying_red_hat_openstack_platform_18.0_development_preview_1_on_red_hat_openshift_container_platform/assembly_creating-the-data-plane_preparing#proc_registering-pre-provisioned-nodes-to-the-data-plane_dataplane


validated_arch_1/stage5/openstackdataplanenodeset.yaml
```diff
@@ -53,8 +53,13 @@ spec:
       ansiblePort: 22
       ansibleUser: cloud-admin
       ansibleVars:
+        # CHANGEME
         edpm_chrony_ntp_servers:
           - clock.redhat.com
+        # CHANGEME -- see https://access.redhat.com/solutions/253273
+        # edpm_bootstrap_command: |
+        #       subscription-manager register --username <subscription_manager_username> --password <subscription_manager_password>
+        #       podman login -u <registry_username> -p <registry_password> registry.redhat.io
```
abays, line 62: Would you mind adding something for these to the `CHANGEME` list at the top as well, like we've done for other `CHANGEME`s?  I think it's good enough to just have it in this file and not to need repeating in the stage6 version, as that version refers back to any customizations made to this version.

abays (approved):
/lgtm

validated_arch_1/stage5/openstackdataplanenodeset.yaml
rabi: There is already a service created in ci-framework and is used in downstream jobs. https://github.com/openstack-k8s-operators/ci-framework/commit/a793f8f18ab8d3b1f8da3e164992f6a26f4c467e

PR #11, opened 2023-10-09
Update VA1 to use Glance image conversion

fultonj:
Add a CR for a PVC where Glance can convert images. Upadate OpenStackControlPlane CR to use the PVC. Also add `enabled_import_methods` with `glance-direct` to `customServiceConfig` sample because a popular way to import an image when conversion is used is with `openstack image create --import` since it forces the use of the import plugin instead of a direct upload.

A qcow2 image which is made into a bootable volume by running:
```
IMG=cirros-0.5.2-x86_64-disk.img
openstack image create my_img --disk-format raw --file $IMG --import
openstack volume create --size 8 my_vol --image $(openstack image show my_img -f value -c id)
```
should show up in the Ceph volumes pool with a reference to its parent in the images pool:
```
[ceph: root@edpm-compute-1 /]# rbd -p volumes ls -l
NAME                                         SIZE   PARENT                                            FMT  PROT  LOCK
volume-5ec39434-9674-4fa8-8ebf-ef0ce8dc22e2  8 GiB  images/d0285227-739d-424c-b4f2-938ee49eb0cc@snap    2
[ceph: root@edpm-compute-1 /]#
```
The above will happen because after the qcow2 file is uploaded, Glance will use the staging area to convert it to a RAW format before import into Ceph. If a volume created from image does not have a parent, then the image was not converted and a full copy was made in the volumes pool.

Jira: OSP-29527

### DIFF
--- a/validated_arch_1/stage6/README.md
+++ b/validated_arch_1/stage6/README.md
@@ -12,22 +12,26 @@ Assumes that a Ceph cluster is available
 ```bash
 oc apply -f ceph_secret.yaml -f nova_ceph.yaml
 ```
-2. Update OpenStackControlPlane and wait for it to finish
+2. Create image conversion PVC
+```bash
+oc apply -f image_conversion_pvc.yaml
+```
+3. Update OpenStackControlPlane and wait for it to finish
 ```bash
 oc apply -f openstackcontrolplane.yaml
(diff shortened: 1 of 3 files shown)


abays (approved):
/lgtm

PR #12, opened 2023-10-12
Simplify HCI deployment with servicesOverride

fultonj:
Use servicesOverride feature of OpenStackDataPlaneDeployment to make HCI example more clear. Instead of having human operator edit the services list before creating a second deployment, provide two deployment examples which can be run before and after Ceph is deployed.

Jira: OSP-29040


abays (approved):
/lgtm

PR #14, opened 2023-10-12
ovn: remove external-ids from the spec

booxter:
These are default values and should apply implicitly since [1].

[1] https://github.com/openstack-k8s-operators/ovn-operator/pull/64

Draft for now since I'd like to first make sure myself this works (if applied via openstackCP, not directly through ovnController CRD.)


abays:
I am going to test this now in a VA1 deployment

abays (approved):
Works in my environment

/lgtm

PR #15, opened 2023-10-13
Drop certain network Ansible vars and rename metadata agent key

abays:
The network vars are automatically calculated now by the data plane operator and the metadata agent key was simply renamed


fultonj (approved):
Thanks @abays

PR #17, opened 2023-10-17
Fix data plane services

abays:
Unfortunately, using the `OpenStackDataPlaneDeployment.spec.servicesOverride` assumes that the actual `OpenStackDataPlaneService` CRs have been created beforehand.  On the other hand, any services include in the `OpenStackDataPlaneNodeSet.spec.services` list are automatically created by the data plane controller itself.  Since we don't want to burden the user with manually creating these, let's switch back to using `OpenStackDataPlaneNodeSet.spec.services` for now.

### DIFF
--- a/validated_arch_1/stage5/openstackdataplanedeployment.yaml
+++ b/validated_arch_1/stage5/openstackdataplanedeployment.yaml
@@ -6,10 +6,6 @@ metadata:
 spec:
   nodeSets:
   - openstack-edpm-ipam
-  servicesOverride:
-    - configure-network
-    - validate-network
-    - install-os
-    - ceph-hci-pre
-    - configure-os
-    - run-os
+  # Listing the OpenStackDataPlaneServices here does work unless they
+  # have been pre-created manually by the user
+  servicesOverride: []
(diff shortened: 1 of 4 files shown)


fultonj:
> Unfortunately, using the `OpenStackDataPlaneDeployment.spec.servicesOverride` assumes that the actual `OpenStackDataPlaneService` CRs have been created beforehand. On the other hand, any services include in the `OpenStackDataPlaneNodeSet.spec.services` list are automatically created by the data plane controller itself. Since we don't want to burden the user with manually creating these, let's switch back to using `OpenStackDataPlaneNodeSet.spec.services` for now.

I'd like to propose an alternative. As per the [docs](https://openstack-k8s-operators.github.io/dataplane-operator/composable_services/#dataplane-operator-provided-services)

"If the services field is omitted from the OpenStackDataPlaneNodeSet spec, then the above list will be used."

Thus, we should be able to remove these two lines:

https://github.com/openstack-k8s-operators/architecture/blob/main/validated_arch_1/stage5/openstackdataplanenodeset.yaml#L156

https://github.com/openstack-k8s-operators/architecture/blob/main/validated_arch_1/stage6/openstackdataplanenodeset.yaml#L141

and user shouldn't have to create the default services (custom services need to be created by user either way).

Could we try that instead so that conceptually the service list can be tied to deployment 1 and deployment 2?

fultonj:
@abays Could we please try this PR instead https://github.com/openstack-k8s-operators/architecture/pull/18 ?

abays:
Closing in favor of https://github.com/openstack-k8s-operators/architecture/pull/18

PR #19, opened 2023-10-17
Define full service list in OpenStackDataPlaneNodeSet

fultonj:
Follow up to ab1a80a0e75e6c555a235d39e1ebf29dd6daa60f

The ceph-hci-pre and ceph-client services will not be created by default so the services list cannot be empty if we want the dataplane operator to define these services for the user. Thus, put all services which will be used in the OpenStackDataPlaneNodeSet so they will be created. We still rely on the servicesOverride in each deployment however to identify which subset of services are deployed when.


abays (approved):
/lgtm

PR #22, opened 2023-10-19
First draft of VA1 with preprovisioned computes

abays:
Removes Metal3 references and changes data plane CRs to use preprovisioned compute nodes


fultonj:
Thank you Andrew

afazekas:
Should we default to metal3 provisioned ?

abays:
> Should we default to metal3 provisioned ?

We were asked by managment and CI to use preprovisioned computes for the first VA due to the technical complexity of using Metal3 in CI.

PR #24, opened 2023-10-24
Rename OpenStackDataPlaneDeployment so it's not deleted

fultonj:
Rename stage5 deloyment from openstack-edpm-ipam to deployment-pre-ceph 
Rename stage6 deloyment from openstack-edpm-ipam to deployment-post-ceph

Do not instruct user to delete their pre-ceph deployment as the logs will be removed. Instead have them create a second deployment. Both can continue to exist after Ansible runs as a record of what deployments have happened.


abays (approved):
/lgtm

PR #27, opened 2023-10-31
Remove image import staging

fultonj:
In the current implemenation this is causing pods to get stuck in pending. Remove until we can implement correctly after the following merges.

https://github.com/openstack-k8s-operators/glance-operator/pull/352


abays (approved):
/lgtm

fmount (approved):

PR #30, opened 2023-11-03
During stage 6 control and data plane are updated

fultonj:
Add note during stage 6 that the control and data plane are updated with `apply -f` and thus any customizations could be lost.


abays (approved):
/lgtm

karelyatin (approved):
/lgtm

Good to have this warning until we have other ways to handle just the update part.

PR #32, opened 2023-11-15
Move VA1 to the new GlanceAPI definition

fmount:
Glance moved to a new API definition via [1][2] and this patch aligns VA1 to the current changes.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/514

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/329
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/514

### DIFF
--- a/validated_arch_1/stage4/openstackcontrolplane.yaml
+++ b/validated_arch_1/stage4/openstackcontrolplane.yaml
@@ -75,10 +75,7 @@ spec:
       route: {}
     template:
       databaseInstance: openstack
-      glanceAPIExternal:
-        networkAttachments:
-          - storage
-      glanceAPIInternal:
+      glanceAPI:
         networkAttachments:
           - storage
         override:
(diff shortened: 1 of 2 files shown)


fultonj (commented):
I'm +2 on this after the depends-on merges

abays (approved):
https://github.com/openstack-k8s-operators/openstack-operator/pull/514 has merged

/lgtm

fmount:
Thanks, I guess we are ready to go here.
However, before merging, I'm wondering if the VA1 is used in combination w/ DP2 content. In that case, shouldn't we need to branch here before merging this change? In theory just a DP2 tag should work as well.
@abays wdyt?

abays:
> Thanks, I guess we are ready to go here. However, before merging, I'm wondering if the VA1 is used in combination w/ DP2 content. In that case, shouldn't we need to branch here before merging this change? In theory just a DP2 tag should work as well. @abays wdyt?

We could do that, but VA1 has never been tested against DP2 even as-is prior to this PR.  My worry would be that a DP2 branch would imply that it has indeed been validated against that.

I think going forward we will need to make specific, focused efforts to test against both upstream and downstream releases and pin branches/tags in the VAs to those.

PR #34, opened 2023-11-16
Restructure for DTs and rename VA1 to HCI

fultonj:
Store not only VAs but also DTs in this repo and create a subdirectory for each to keep them separate.

Move `validated_arch_1` to the new `va` subdirectory and rename it `hci`.

Create `dt` subdirectory and include an explanation on how a DT differs from a VA.


fultonj:
Let's hold until we have agreement. Preview here https://github.com/fultonj/architecture/tree/dt_va_sub

abays (approved):
/lgtm

fmount (approved):
+1 from me

ASBishop:
+1 from me as well

PR #36, opened 2023-11-29
Clarify use of pre-provisioned computes

abays:
Removes a reference to Metal3 provisioning and directly notes the use of pre-provisioned computes instead

### DIFF
--- a/va/hci/stage1/README.md
+++ b/va/hci/stage1/README.md
@@ -4,7 +4,7 @@ Install the dependencies for Red Hat OpenStack Services on OpenShift operators
 
 ## Notes
 
-Requires an OpenShift 4.12+ cluster with 3 master/worker combo-nodes, with Metal3 and a provisioning network available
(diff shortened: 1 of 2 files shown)


fultonj (approved):

va/hci/stage1/README.md
```diff
@@ -4,7 +4,7 @@ Install the dependencies for Red Hat OpenStack Services on OpenShift operators
 
 ## Notes
 
-Requires an OpenShift 4.12+ cluster with 3 master/worker combo-nodes, with Metal3 and a provisioning network available
+Requires an OpenShift 4.12+ cluster with 3 master/worker combo-nodes
```
SeanMooney, line 7: this is still outdated by the way
we should be using 4.14+ now.

4.14 was released on 01 Nov 2023 and should be the min version we test with going forward.

can you update this in a future pr

PR #37, opened 2023-12-01
Introduce common directory

fultonj:
The common directory prevents duplicate CR files which all VAs and DTs are expected to need as explained in common/README.md.

- Move all of va/hci/stage{1,2,3} to common common/stage{1,2,3} 
- Move select files from stage{4,5,6} to common/stage{4,5,6} 
- Add symbolic links to all moved files.


fultonj:
> +1 from me, but I want to raise a question.
> 
> Instead of adding symlinks in each va/dt stage to the corresponding file in the common area, we could update the README's in each stage to "oc apply" the common file. That would help emphasize to the reader that they're applying a common CR. It would eliminate the broken link problem, but I suppose it would replace it with a corresponding "broken documentation" problem.

Yes, I like the idea of having only one copy of certain common files and then each VA/DT only has separate directories where they actually differ. E.g. Have only one OLM directory which every VA/DT will use and then each VA/DT can have pre-requisites which are that certain common directories be applied. I can try that in another PR and put this one in a state of draft for now.

fultonj:
> > +1 from me, but I want to raise a question.
> > Instead of adding symlinks in each va/dt stage to the corresponding file in the common area, we could update the README's in each stage to "oc apply" the common file. That would help emphasize to the reader that they're applying a common CR. It would eliminate the broken link problem, but I suppose it would replace it with a corresponding "broken documentation" problem.
> 
> Yes, I like the idea of having only one copy of certain common files and then each VA/DT only has separate directories where they actually differ. E.g. Have only one OLM directory which every VA/DT will use and then each VA/DT can have pre-requisites which are that certain common directories be applied. I can try that in another PR and put this one in a state of draft for now.

I'm trying an alternative approach which uses `kustomize` 
[components](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/1802-kustomize-components#summary) to solve the duplication problem.

Still in draft https://github.com/openstack-k8s-operators/architecture/pull/39

fultonj:
superseded by https://github.com/openstack-k8s-operators/architecture/pull/39

PR #28, opened 2023-10-31
Refactor to use kustomize with bases

mdbooth:
This primarily:
* Moves and renames stages 1-4
* Defines stages 1-4 as kustomize resources
* Composes stages 1 and 2 as a single 'olm' resource
* Composes stages 3 and 4 as a single 'osp-control-plane' resource

Stages 5 and 6 become edpm-stage1 and edpm-stage2, but they are otherwise unchanged.

Additionally:
* nmstate_deploy and metallb_deploy move out of the former stage1 into 'networking'
* osp-secrets is defined from a plaintext envfile

Otherwise it makes no functional changes.

PR #38, opened 2023-12-07
Adding DT template

EliadCohen:
Included a modified copy of the first VA as a template/skeleton Added a Makefile and rendering of the template name, directory and date of first revision, all from a render.sh script


EliadCohen:
@fultonj What do you think?

fultonj:
> Thanks @fultonj , that makes sense. I want to improve this PR, if you have a minute to convey some thoughts I'd love to hear. 

I'm working on something which will combine the ideas from these two PRs (which will probably stay in draft).

- https://github.com/openstack-k8s-operators/architecture/pull/37
- https://github.com/openstack-k8s-operators/architecture/pull/28

> WRT the duplication, yes. I'm with you. 

Solving duplication is hard but important. Ideas from PR28 are one way to solve it. I'll add you to my PR as soon as it's working.

> Are the stages rigid in the definition of what happens during them? I'm just not sure how this is being consumed on the other end

The stages are not rigid. I'm looking to replace them with semantic names. It will be more clear when you see the PR. I want to get it working first since it's a big change and will be a lot to take in so I want to refine it first.

fultonj:
> @fultonj it is now rebased. 

I meant it should be rebased on the PR I have not yet sent in. That PR should help with the duplication problem so the files changed will be much smaller.

>  Do you think we should replace links with files when we copy the template of the dt? 
> Or keep links unless we need to modify an individual file?

If we're going to use links, then we would have lots of links and only have individual files on as needed basis.

However, I'd rather use neither of the above two options and instead use kustomize so solve the duplication problem. How that works will be clear when I send in my PR. 

> and also a question on the keys with `# CHANGEME` values. Do we expect to do the replacements during execution?

Right now the problem is that people have to change the `# CHANGEME` values. When the new PR comes kustomize will replace them automatically. An example of how that works is here:

https://github.com/openstack-k8s-operators/architecture/pull/28

fultonj:
I'd prefer if this was rebased on this:

   https://github.com/openstack-k8s-operators/architecture/pull/39

Note that PR39 is under heavy development still

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
+# RHOS Deployed Topology %NEW_DT_NAME%
```
tosky, line 1: I would remove RHOS as it's not so relevant as a DT can be technically used without.

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+**Based on OpenStack K8S operators from the "main" branch of the [OpenStack Operator repo](https://github.com/openstack-k8s-operators/openstack-operator/tree/78b3c876eaf9168f9d95b201997ebdc2da42fa02) on Oct 17th, 2023**
```
tosky, line 3: The commit should be enought, I would remove the date.

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## General information
+
+| Revision | Change                | Date             |
+|--------: | :-------------------- | :--------------: |
+| v0.1     | Initial publication   | %TODAY%      |
+
```
tosky, line 10: I would add here an optional purpose section to explain the idea behind the DT

## Purpose

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+| FIPS Mode                                   | default                         | Must have/standard |
+
+## Considerations
+
+1. Glance requirement dictates specific lab
+2. HSM access limited geographically
+
+## Testing tree
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
```
tosky, line 38: "Test case to report" is not needed: it may be a list too big, it may be just changed into a reference to the test run, but as it is not a public information it should not live here

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Node topology
+| Node role                                     | bm/vm | amount |
+| --------------------------------------------- | ----- | ------ |
+| Openshift master/worker combo-node cluster    | vm    | 3      |
+| Compute nodes                                 | vm    | 3      |
+| Galera DB node, replicating                   | vm    | 3      |
+| Hardware security module                      | bm    | 1      |
+
+## Services, enabled features and configurations
```
tosky, line 19: I would split the table into two: 
- the table like the current one with the list of services which are focus of the deployment and testing of this DT
-  a separate table with just two columns (service and reason) which lists the additional services that needs to be deployed (typically keystone and co) which are not the focus of the testing but they are needed for the full integration testing to happen, under a "### Support services" section

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+| ------------------------------------------- | ------------------------------- | ------------------ |
+| RabbitMQ                                    | default                         | Must have          |
+| OVN                                         | default                         | Must have          |
+| RabbitMQ                                    | default                         | Must have          |
+| Cinder                                      | Ceph/RGW backend                | optional           |
+| Manila                                      | Ceph/Ganesha+MDS Backend        | Must have          |
+| Glance                                      | NetApp, Fibre channel           | Must have          |
+| Keystone                                    | Special appliance [redacted]    | Interchangable     |
+| FIPS Mode                                   | default                         | Must have/standard |
+
```
tosky, line 30: I would add an additional optional subsection ("### Additional configuration") to list "default" stuff (like TLSe and FIPS, just for reminder).

templates/dt_template/README.md
```diff
@@ -0,0 +1,58 @@
...
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/compute  | stage7       | Use rhel image        | 11223344            |
+| Tempest/scenairo | stage7       | Use cirros image      | 22334455            |
+| Jordan           | stage7       | None                  | 44556677            |
+| Rally/Browbeat   | stage8       | Use CS9 image         | 55667788            |
+| Tobiko/Faults    | stage9       | Use cirros image      | 33445566            |
+
+## Stages
```
tosky, line 46: I would make this optional, especially if the stages are going to be "standardized"

fultonj: I think it's OK to think in terms of stages for now. In the long run I don't think they'll be standardized. This can be rebased if we switch to kustomize as seen here.

https://github.com/openstack-k8s-operators/architecture/pull/39

templates/dt_template/README.md
EliadCohen: Done

templates/dt_template/README.md
EliadCohen: How do we capture the info regarding which coverage needs for which team this addresses?

templates/dt_template/README.md
EliadCohen: Done, good idea

templates/dt_template/README.md
EliadCohen: Done

templates/dt_template/README.md
tosky: My point was about having non-numerical names as in your patch, that's a way to standardize them (once you have a stage called "control-plane" of course people won't duplicate that, but improve it)

templates/dt_template/README.md
tosky: Here  - or better, i nthe code which implements the PR - you could see the list of tests executed (if tempest, the path of the tempest tests; if tobiko, same; if horizon scenario tests, again the list of tests or all).

templates/dt_template/README.md
```diff
@@ -0,0 +1,80 @@
+# Deployed Topology 
+
+**Based on OpenStack K8S operators from the "main" branch of the [OpenStack Operator repo](https://github.com/openstack-k8s-operators/openstack-operator/tree/78b3c876eaf9168f9d95b201997ebdc2da42fa02) on Oct 17th, 2023**
+
+## General information
+
+| Revision | Change                | Date             |
```
tosky, line 7: Thinking more about this, do we really need this table? We use git, we can trace the history and I fear this table will become quickly outdated.

templates/dt_template/README.md
```diff
@@ -8,6 +8,10 @@
 |--------: | :-------------------- | :--------------: |
```
tosky, line 8: I thought I left a message also about this, but do we need the table? Can't we rely on the git history? This is going to quickly rot.

tosky:
This should be rebased too

PR #39, opened 2023-12-11
Restructure for kustomize

fultonj:
This patch solves the following problems by using kustomize:

A. Duplication: A new VA/DT would result in duplicate files
B. Substitution: User must update every line with a #CHANGEME

kustomize inheritance is insufficient to solve these problems so this patch uses components [1]. The original CRs are moved to a shared library directory with new names:

- stage1 to olm-deps
- stage2 to olm-osp
- stage3 to networking
- stage4 to control-plane

The duplication problem is solved by having each VA/DT use a kustomization file which composes only the required CRs from the common library. A common/ directory is also introduced for CRs which all VAs/DTs are expected to need (e.g. OLM).

The substitution problem is solved by using `replacements` (see `lib/control-plane/kustomization.yaml`) from a user-provided file (see `examples/va/hci/values.yaml`).

The new examples/ directory becomes the user's main interface and the following generates the CRs from stages 1-4:
```
kustomize build examples/common/olm/
kustomize build examples/va/hci/
```
In `examples/va/hci/kustomization.yaml` a URL is used to reference a copy of this repository so that this repository can be the upstream of a user's own kustomize pipeline.

Note: stage5/ and stage6/ are renamed to edpm-pre-ceph/ and edpm-post-ceph/ respectively but are left in the user-facing hci/ directory since they are not yet migrated to kustomize. Instructions are updated for the new location but not yet in how they are used.

This is based on work [2] originally proposed by Matthew Booth.

[1] https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/1802-kustomize-components
[2] https://github.com/openstack-k8s-operators/architecture/pull/28


examples/va/hci/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Considerations
+
+1. These CRs are validated for the overall functionality of the OSP cloud deployed, but they nonetheless require customization for the particular environment in which they are utilized.  In this sense they are _templates_ meant to be consumed and tweaked to fit the specific constraints of the hardware available.
+
+2. The CRs are applied against an OpenShift cluster in _stages_.  That is, there is an ordering in which each grouping of CRs is fed to the cluster.  It is _not_ a case of simply taking all CRs from all stages and applying them all at once.
+
+3. In stages 1 and 2 [kustomize](https://kustomize.io/) is used to genereate the CRs dynamically. The `values.yaml` file(s) must be updated to fit your environment. kustomize version 5 or newer required.
+
+4. In stages 3 and 4 YAML comments are placed throughout the CRs to aid in the process of customizing the CRs.  Fields that _must_ (or most likely need to be) changed are commented with "# CHANGEME" either on the field itself or somewhere nearby.  Other comments are added to explain fields that can be changed and, sometimes, to explain additions that can be made.
```
abays, line 28: Should we remove the mention of `# CHANGEME` here, since we're using `kustomize` for all of that now?  Or are there still places where we rely on `# CHANGEME` for particular reasons?

fultonj: Stages 3 and 4 still have `#CHANGEME` because they have not yet been moved to kustomize. They refer to `edpm-pre-ceph` and `edpm-post-ceph`.

I was thinking of moving them to kustomize in a follow up PR since this change is pretty big.

lib/networking/kustomization.yaml
```diff
@@ -0,0 +1,367 @@
...
+    kind: ConfigMap
+    name: network-values
+    fieldPath: data.ctlplane.iface
+  targets:
+  - select:
+      group: metallb.io
+      kind: L2Advertisement
+      name: ctlplane
+    fieldPaths:
+    - spec.interfaces.0
```
rebtoor, line 327: I've opened a PR against your branch to add other interfaces: https://github.com/fultonj/architecture/pull/1/files

rebtoor (commented):
I reviewed this patch following your comment [1](https://github.com/openstack-k8s-operators/architecture/pull/39#issuecomment-1855795505). I think it's almost ok, i opened a PR against your branch for some missing interface configurations in the `L2Advertisement` CRs and a clarification request.

lib/networking/kustomization.yaml
```diff
@@ -0,0 +1,367 @@
...
+- netattach_ctlplane.yaml
+- netattach_internalapi.yaml
+- netattach_storage.yaml
+- netattach_tenant.yaml
+- netconfig.yaml
+
+patches:
+- target:
+    kind: NodeNetworkConfigurationPolicy
+    labelSelector: "osp/nncm-config-type=standard"
```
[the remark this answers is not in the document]
fultonj: This is intended. 

We originally added it so people could run this during [stage 3](https://github.com/openstack-k8s-operators/architecture/tree/main/va/hci/stage3).

```
oc wait nncp -l osp/interface=enp7s0 --for condition=available --timeout=300s
```
Of course the interface name could change though and we'd have to have kustomize change labels. So instead we just have have a `labelSelector` for the "standard NNCM-config" so that regardless of name people can just run this:
```
oc wait nncp -l osp/nncm-config-type=standard --for condition=available --timeout=300s
```
I updated the docs too (hci.md for now).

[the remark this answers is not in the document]
rebtoor: Thanks for the explanation @fultonj !

lib/networking/metallb_l2advertisement.yaml
```diff
@@ -12,8 +12,7 @@ spec:
   ipAddressPools:
   - ctlplane
   interfaces:
-  # Change this interface to the appropriate OpenStack NIC for your environment
-  - enp7s0 # CHANGEME
+  - _repalced_
```
gouthampacha, line 15: typo? here and below

fultonj: This is not a typo. 

Values like `enp7s0` should not be in the `lib` directory. They go in the values.yaml file under `examples`. So where those values were in the `lib` directory, I put a dummy value like `_replaced_`. Then when kustomize is run, the rendered output has the value and the dummy value is replaced.

gouthampacha: yes; i meant there's a real typo in here :)

fultonj: Oh! I didn't see it because I read `replaced` too quickly. You meant "you spelled 'replaced' wrong".
```
-  - _repalced_
+  - _replaced_
```

examples/va/hci/values-ci-framework.yaml
```diff
@@ -0,0 +1,205 @@
...
+          - end: 192.168.122.120
+            start: 192.168.122.100
+          - end: 192.168.122.200
+            start: 192.168.122.150
+        cidr: 192.168.122.0/24
+        gateway: 192.168.122.1
+        name: subnet1
+    prefix-length: 24
+    iface: enp6s0
+    mtu: 1500
```
abays, line 52: I had to change this to `9000` in order for the NNCPs to successfully configure the interfaces for `storage` and `storageMgmt`.  Otherwise I saw this error in the NMState logs on the OCP nodes:

```
NmstateError: VerificationError: Verification failure: storage.interface.mtu desire '9000', current '1500'
```

I think that's because the base interface has to have MTU >= the MTU you desire to set on any VLAN interface attached to it.

fultonj: Good catch. I've updated values.yaml so the interface has 9000.

Also, I changed the mtu values in lib from _replaced_ to 1500 so that they are substituted as integers, not strings.

examples/va/hci/control-plane.md
```diff
@@ -0,0 +1,66 @@
...
+Change to the hci directory
+```
+cd architecture/examples/va/hci
+```
+Edit the [values.yaml](values.yaml) file to suit your environment.
+```
+vi values.yaml
+```
+Alternatively use your own copy of `values.yaml` and edit 
+[kustomization.yaml](kustomization.yaml) to use a that copy.
```
abays, line 22: `s/a that/that/`

fultonj: completed

SeanMooney:
this is not a full review but overall this looks quite good to me.
i still need to fully exaime the generated output and try this myself but i can see the outline of how i would extend this for other DTs/VAs

so +1 from me on moving forward with this so that we can start working on DTs that leverage this common framework

PR #46, opened 2023-12-14
Security DFG DT commit

jagee:


tosky:
My question is: do all of these need to be DTs? Can't those be run at the component level? Especially the ones which use specific backends. It is true they may need a few more components, but generally speaking if you enable image signature you will be just able to run a limited amount of tests, which makes it suitable for running at that part of the pipeline.

millevy:
> My question is: do all of these need to be DTs? Can't those be run at the component level? Especially the ones which use specific backends. It is true they may need a few more components, but generally speaking if you enable image signature you will be just able to run a limited amount of tests, which makes it suitable for running at that part of the pipeline.

Hi, actually we do need to get most of it deleted and move to Jiras. I was trying to figure out how to. I will move it now to Jiras and I will ask Jeremy to change his PR.

millevy (changes requested):
This DT belong to the component ci and need to be removed.

millevy (changes requested):
Please remove all the DTs that belong to the Component CI

PR #57, opened 2023-12-20
Wait nncp for SuccessfullyConfigured condition

karelyatin:
It's more strict as condition=available matches
even for NoMatchingNode.


abays (approved):
/lgtm

PR #58, opened 2023-12-20
Fix Glance internal service network override

abays:
We have changed the network override structure for Glance, which we need to reflect in our `OpenStackControlPlane` CRs.


fmount (approved):
/lgtm

fmount:
@eduolivares FYI this should fix your deployment as well! Thanks @abays for this patch!

PR #60, opened 2023-12-22
Add compute-starter-kit dt

SeanMooney:
This change adds a redme describign the comptue starter kit
DT and job variants. This DT may be graducated to a VA in the future
and will be used as the basis of the compute content promotion jobs
prior to integration of promoted content.


SeanMooney:
@fultonj i have not actually tested this beyond rendering the crs with kustomize locally but this is a port of the old nova compute kit ci toplogy i wrote last year overlayed on the lib functions you intoduced

https://github.com/openstack-k8s-operators/nova-operator/tree/main/ci/nova-operator-compute-kit

one slight annoyance is ideaaly we would start with a contol plane with all service disabled and then enabled only the ones needed 
to work around that i have used patches to remvoe all service form the default template provided in teh contolplane lib that are not needed and then modifed  the remaining services as need.

i have a couple of opens with the current version mainly around how do i correctly expose the rest apis as routes?

and then obviously i need to extend this to have 2 edpm nodes in cell 1 before it actully complete

examples/dt/compute/compute-starter-kit/README.md
```diff
@@ -0,0 +1,119 @@
+# Compute Starter Kit
+
+**Based on OpenStack K8S operators from the "main" branch of the [OpenStack Operator repo](https://github.com/openstack-k8s-operators/openstack-operator/)**
+
+This is a collection of CR templates that represent a openstack deployment
+topology that has the following characteristics:
+
+- Single noe OpenShift cluster (CRC)
```
fultonj, line 8: node

examples/dt/compute/compute-starter-kit/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+3. In stages 1 and 2 [kustomize](https://kustomize.io/) is used to genereate the CRs dynamically. The `values.yaml` file(s) must be updated to fit your environment. kustomize version 5 or newer required.
+
+4. In stage 3 YAML comments are placed throughout the CRs to aid in the process of customizing the CRs.  Fields that _must_ (or most likely need to be) changed are commented with "# CHANGEME" either on the field itself or somewhere nearby.  Other comments are added to explain fields that can be changed and, sometimes, to explain additions that can be made.
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install the OpenStack K8S operators and their dependencies]../(../../common/)
```
fultonj, line 117: Please correct the markdown link syntax so it renders correcty.

[Install the OpenStack K8S operators and their dependencies](../../../common/)

fultonj (commented):
Looks promising. This is just a first pass. I'll do a follow up review.

examples/dt/compute/compute-starter-kit/control-plane.md
```diff
@@ -0,0 +1,66 @@
...
+
+- A storage class called `crc-csi-hostpath-provisioner` should already exist.
+
+## Initialize
+
+Switch to the "openstack" namespace
+```
+oc project openstack
+```
+Change to the hci directory
```
jamepark4, line 13: Wouldn't this be change to the compute-starter-kit directory?

examples/dt/compute/compute-starter-kit/values.yaml
```diff
@@ -0,0 +1,189 @@
...
+    endpoint_annotations:
+      metallb.universe.tf/address-pool: internalapi
+      metallb.universe.tf/loadBalancerIPs: 172.17.0.85
+  rabbitmq-cell1:
+    endpoint_annotations:
+      metallb.universe.tf/address-pool: internalapi
+      metallb.universe.tf/loadBalancerIPs: 172.17.0.86
+
+  lbServiceType: LoadBalancer
+  storageClass: crc-csi-hostpath-provisioner
```
gibizer, line 189: Can we define additional nova compute configuration and then use customize to copy that into the DataPlaneService/nova? E.g. enabling cpu pinning for the second scenario

SeanMooney: proably yes i have not got to the edpm part yet.

SeanMooney:
this is just a rebase to take account of the changes in the repo and i fixed 2 issues
rabbit now requests enouch memory to avoid being OOM killed and i set the database user names for nova.

with this nova/placment deploy properly

glance is failing with some pod creation issue related to the storage network we will have to figure that out. my guess is the network config in this dt is not compatible with CRC currently and needs to be adjusted but i have not really had time to figure out how that is working

examples/dt/compute/compute-starter-kit/values.yaml
```diff
@@ -0,0 +1,189 @@
...
+          - end: 172.18.0.250
+            start: 172.18.0.100
+        cidr: 172.18.0.0/24
+        name: subnet1
+        vlan: 21
+    mtu: 9000
+    prefix-length: 24
+    iface: enp7s0.21
+    vlan: 21
+    base_iface: enp7s0
```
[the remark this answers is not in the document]
SeanMooney: yep for crc this is

enp6s0

examples/dt/compute/compute-starter-kit/control-plane.md
SeanMooney: yep
this file is mostly just a copy past right now but ill correct this when i get to the datapalne part

this dt will have at least the following phases.

deploy openstack with nova and cell0
deploy cell1
 - first deploy the contolplane part of cell1 and then waith for ti to be readay
 - then deploy the dataplane
finally perfrom post install config (flavor/image creations) and run tests.

examples/dt/compute/compute-starter-kit/values.yaml
```diff
@@ -0,0 +1,189 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: network-values
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  # nodes
+  node_0:
+    name: crc
```
SeanMooney, line 11: ok this needs to be updated to match  the name of the CRC VM

for my current instance this should be 

 node_0:
    name: crc-n5gv4-master-0
    internalapi_ip: 172.17.0.5
    tenant_ip: 172.19.0.5
    ctlplane_ip: 192.168.122.19
    storage_ip: 172.18.0.5

that is why the network config was broken
it was not finding a node called crc

examples/dt/compute/compute-starter-kit/kustomization.yaml
```diff
@@ -0,0 +1,53 @@
...
+      path: /spec/swift
+    # - op: remove
+    #   path: /spec/barbican
+  target:
+    kind: OpenStackControlPlane
+    name: controlplane
+    version: v1beta1
+- path: control-plane/disable.yaml
+- path: control-plane/infra.yaml
+- path: control-plane/nova.yaml
```
SeanMooney, line 53: glance and neutron seam to be defaulting to 0 replicas so i just need to add a patch file for each and they should deploy

leifmadsen:
> to me this is required before we can move the ci framework jobs to all be based on the architecture repo.
> 
> so this is not a priority for beta or even ga but i actually think this should be elevated to a VA i have just been two busy to work on this due to other work

Perhaps this should be moved to a Jira for tracking, scoping, and backlog prioritization rather than leaving this hanging open. If this is not a target for beta or GA, and is more tied to CIFMW integration, then I think it might be best to offline this and start with planning what needs to be done.

SeanMooney:
orignally this was motaivate by the orginal request for DT which was also communicated as applying to the component piplien and github jobs.

later that request was removed but the desire to have a minimal configuration for development and ci
i.e. that works with a single openshfit worker node (CRC, microshfit, SNO) 

this stopped being a ga requirement when having dt for all jobs including upstream github and downstream component pipepliens was removed a few months ago which is why i depriotised this.

PR #61, opened 2024-01-03
Add DT for Director Operator Adaption to Podified

pinikomarov:


dt/ospdo_dt.md
```diff
@@ -0,0 +1,43 @@
+    | vm    | 3      |
```
psathyan, line 1: Would you need compact Openshift source and destination?
Manila is configured with `ceph`, I believe you would need 3 additional nodes for Ceph?

dt/ospdo_dt.md
```diff
@@ -0,0 +1,43 @@
...
+| Tempest/scenairo | stage7       | Use cirros image      | 22334455            |
+| Jordan           | stage7       | None                  | 44556677            |
+| Rally/Browbeat   | stage8       | Use CS9 image         | 55667788            |
+| Tobiko/Faults    | stage9       | Use cirros image      | 33445566            |
+
+## Stages
+
+All stages must be executed in the order listed below.  Everything is required unless otherwise indicated.
+
+1. Deploy a Director Operator env
```
psathyan, line 32: Would this step be done outside of `ci-framework`?

psathyan (commented):
Some queries.

dt/ospdo_dt.md
```diff
@@ -0,0 +1,43 @@
...
+8. Execute OSP pre deployment per DFG specific testing
+2. Collect services' DB from OSPdO env
+3. Prepare second env for Podified Deploymnt
+4. Install dependencies for the OpenStack K8S operators
+5. Migrate DB to Podified env's Operators
+6. Install the OpenStack K8S operators
+7. Configuring networking on the OCP nodes
+8. Configure and deploy the control plane
+7. Execute OSP post adaption sanity testing
+8. Execute OSP post deployment Integrate with DFG specific testing as needed
```
fultonj, line 43: If these are numerical steps do you want to correct the numbering? Or perhaps just use bullets?

pinikomarov: done

dt/ospdo_dt.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Services, enabled features and configurations
+| Service                                     | configuration                   | Lock-in coverage?  |
+| ------------------------------------------- | ------------------------------- | ------------------ |
+| RabbitMQ                                    | default                         | Must have          |
+| OVN                                         | default                         | Must have          |
+| RabbitMQ                                    | default                         | Must have          |
+| Cinder                                      | Ceph/RGW backend                | optional           |
+| Manila                                      | Ceph/Ganesha+MDS Backend        | Must have          |
+| Glance                                      | NetApp, Fibre channel           | Must have          |
```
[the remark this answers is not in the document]
pinikomarov: That's is part of my intention , 
Were is the rerence files for VA1 - the base for 17.1 ? 
I mean Where are they her ein the Archutecture repo ? not the gitlab snippet (https://gitlab.cee.redhat.com/-/snippets/7501)
And how can I mention that this DT is a subset of the Base VA1 example ?
Becase I don't think each DT should be presented differently if they stem from the same branch then they should only show the base and differences from the Base architecture no ?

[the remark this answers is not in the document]
tosky: The reference files for VA1 are
https://github.com/openstack-k8s-operators/architecture/tree/main/examples/va/hci

What you have here in these table is not VA1. Here you are definining something which uses a combination of backends - as you have ceph around, it makes sense to configure all services with backends to use ceph (this is what VA1 does for example).

PR #62, opened 2024-01-10
SRIOV VA

abays:
The first iteration of the SRIOV implementation for the VA2 effort

### DIFF
--- a/examples/va/hci/kustomization.yaml
+++ b/examples/va/hci/kustomization.yaml
@@ -7,5 +7,54 @@ components:
   ## It's possible to replace ../../../va/hci/ with a git checkout URL as per:
   ## https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
 
+# Add StorageMgmt network template, as it is needed for CephHCI
+patches:
+- target:
+    version: v1beta1
+    kind: NetConfig
+    name: netconfig
+  patch: |-
+    - op: add
+      path: /spec/networks/-
+      value:
+        dnsDomain: _replaced_
+        name: StorageMgmt
+        subnets:
+          - _replaced_
+        mtu: 1500
+
 resources:
   - values.yaml
+
+# Add StorageMgmt network replacements
+replacements:
+# NetConfig dnsDomain specific to this VA
+- source:
+    kind: ConfigMap
+    name: network-values
+    fieldPath: data.storagemgmt.dnsDomain
+  targets:
+  - select:
+      kind: NetConfig
+    fieldPaths:
+    - spec.networks.[name=StorageMgmt].dnsDomain
+# NetConfig MTU specific to this VA
+- source:
+    kind: ConfigMap
+    name: network-values
+    fieldPath: data.storagemgmt.mtu
+  targets:
+  - select:
+      kind: NetConfig
+    fieldPaths:
+    - spec.networks.[name=StorageMgmt].mtu
+# NetConfig subnets specific to this VA
+- source:
+    kind: ConfigMap
+    name: network-values
+    fieldPath: data.storagemgmt.subnets
+  targets:
+  - select:
+      kind: NetConfig
+    fieldPaths:
+    - spec.networks.[name=StorageMgmt].subnets
--- a/examples/va/nfv/sriov/README.md
+++ b/examples/va/nfv/sriov/README.md
@@ -1,0 +1,30 @@
+# OpenStack SRIOV
+
+**Based on OpenStack K8S operators from the "main" branch of the [OpenStack Operator repo](https://github.com/openstack-k8s-operators/openstack-operator/tree/12c57baeca4ae33dd30a7707d330eb094309b4cd) on Dec 6th, 2023**
+
+This is a collection of CR templates that represent a validated Red Hat OpenStack Services on OpenShift deployment that has the following characteristics:
+
+- 3 master/worker combo-node OpenShift cluster
+- 3-replica Galera database
+- RabbitMQ
+- OVN networking
+- Network isolation over a single NIC
+- 2 compute nodes with an additional SRIOV-enabled NIC
+
+## Considerations
+
+1. These CRs are validated for the overall functionality of the OSP cloud deployed, but they nonetheless require customization for the particular environment in which they are utilized.  In this sense they are _templates_ meant to be consumed and tweaked to fit the specific constraints of the hardware available.
+
+2. The CRs are applied against an OpenShift cluster in _stages_.  That is, there is an ordering in which each grouping of CRs is fed to the cluster.  It is _not_ a case of simply taking all CRs from all stages and applying them all at once.
+
+3. In stages 1 and 2 [kustomize](https://kustomize.io/) is used to genereate the CRs dynamically. The `*-values.yaml` file(s) must be updated to fit your environment. kustomize version 5 or newer required.
+
+4. In stage 3 YAML comments are placed throughout the CRs to aid in the process of customizing the CRs.  Fields that _must_ (or most likely need to be) changed are commented with "# CHANGEME" either on the field itself or somewhere nearby.  Other comments are added to explain fields that can be changed and, sometimes, to explain additions that can be made.
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install the OpenStack K8S operators and their dependencies](../../../common/)
+2. [Configuring networking and deploy the OpenStack control plane](control-plane.md)
+3. [Configure and deploy the data plane](edpm)
--- a/examples/va/nfv/sriov/control-plane.md
+++ b/examples/va/nfv/sriov/control-plane.md
@@ -1,0 +1,70 @@
+# Configuring networking and deploy the OpenStack control plane
+
+## Assumptions
+
+- A storage class called `local-storage` should already exist.
+
+## Initialize
+
+Switch to the "openstack" namespace
+```
+oc project openstack
+```
+Change to the nfv/sriov directory
+```
+cd architecture/examples/va/nfv/sriov
+```
+Edit the [values.yaml](values.yaml) and
+[service-values.yaml](service-values.yaml) files to suit 
+your environment.
+```
+vi values.yaml
+vi service-values.yaml
+```
+Alternatively use your own copies of those files and edit
+[kustomization.yaml](kustomization.yaml) to use those copies.
+```
+resources:
+  - values-ci-framework.yaml
+  - service-values-ci-framework.yaml
+```
+
+Generate the control-plane and networking CRs.
+```
+kustomize build > control-plane.yaml
+```
+
+## Create CRs
+```
+oc apply -f control-plane.yaml
+```
+
+Wait for NNCPs to be available
+```
+oc wait nncp -l osp/nncm-config-type=standard --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured --timeout=300s
+```
+
+Wait for control plane to be available
+```
+oc wait osctlplane controlplane --for condition=Ready --timeout=600s
+```
+
+## Workaround
+
+The `control-plane.yaml` file contains CRs for both `NMState` and
+`NodeNetworkConfigurationPolicy` (NNCP). When `oc apply -f` is
+passed this file, OpenShift might try to create the NNCPs while
+`NMState` CRDs are still installing and produce the following message.
+
+```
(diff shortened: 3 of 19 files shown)


examples/va/nfv/sriov/control-plane.md
```diff
@@ -0,0 +1,70 @@
...
+cd architecture/examples/va/nfv/sriov
+```
+Edit the [network-values.yaml](network-values.yaml) and
+[service-values.yaml](service-values.yaml) files to suit 
+your environment.
+```
+vi network-values.yaml
+vi service-values.yaml
+```
+Alternatively use your own copyies of those files and edit
```
fultonj, line 24: copies

examples/va/nfv/sriov/control-plane.md
```diff
@@ -0,0 +1,70 @@
...
+```
+oc wait nncp -l osp/nncm-config-type=standard --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured --timeout=300s
+```
+
+Wait for control plane to be available
+```
+oc wait osctlplane controlplane --for condition=Ready --timeout=600s
+```
+
+## Workaround
```
fultonj, line 52: Perhaps the common directory should have a markdown file describing individual workarounds and both this readme and the HCI read me could link to it. That would avoid content duplication.

examples/va/hci/kustomization.yaml
```diff
@@ -7,5 +7,54 @@ components:
   ## It's possible to replace ../../../va/hci/ with a git checkout URL as per:
   ## https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
 
+# Add StorageMgmt network template, as it is needed for CephHCI
+patchesJson6902:
```
fultonj, line 11: This looks like a good fix to the problem of storage-mgmt being on EDPM nodes only when HCI is used. I have to try it in my env but it looks like the right idea.

examples/va/nfv/sriov/kustomization.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - ../../../../va/nfv/sriov/
+  # - https://github.com/openstack-k8s-operators/architecture/va/hci?ref=main
```
fultonj, line 6: The URL should be `https://github.com/openstack-k8s-operators/architecture/va/nfv/sriov?ref=main`

Alternatively, comments on line 6 to 8 should be removed.

examples/va/nfv/sriov/service-values.yaml
```diff
@@ -0,0 +1,20 @@
+# local-config: referenced, but not emitted by kustomize
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: service-values
```
fultonj, line 5: I like having a separate service-values.yaml file.

examples/va/nfv/sriov/network-values.yaml
```diff
@@ -0,0 +1,191 @@
...
+    endpoint_annotations:
+      metallb.universe.tf/address-pool: internalapi
+      metallb.universe.tf/loadBalancerIPs: 172.17.0.85
+  rabbitmq-cell1:
+    endpoint_annotations:
+      metallb.universe.tf/address-pool: internalapi
+      metallb.universe.tf/loadBalancerIPs: 172.17.0.86
+
+  lbServiceType: LoadBalancer
+  storageClass: host-nfs-storageclass
```
fultonj, line 191: The majority of these values are network-values so the name makes sense. Technically storageClass is separate so we could have storage-values but it would only have one item right now. Maybe we should call it `general-values` instead of `network-values` though.

lib/networking/kustomization.yaml
```diff
@@ -242,15 +242,6 @@ replacements:
       kind: NetConfig
     fieldPaths:
     - spec.networks.[name=Storage].dnsDomain
-- source:
```
fultonj, line 245: ack, I assume this is added back in the output via the update in va/hci/kustomization.yaml

lib/networking/kustomization.yaml
```diff
@@ -242,15 +242,6 @@ replacements:
...
-    kind: ConfigMap
```
fultonj, line 246: ditto

fultonj (commented):
Asking for a few small changes, though overall this looks really good. Thanks Andrew.

examples/va/hci/kustomization.yaml
fultonj, line 11: FYI, though maybe you don't get the deprecation warning with 5.0?

```
[fultonj@jove hci{va2_sriov}]$ kustomize build . > ~/va2.yaml
# Warning: 'patchesJson6902' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
[fultonj@jove hci{va2_sriov}]$ 

[fultonj@jove hci{va2_sriov}]$ kustomize version
v5.1.0
[fultonj@jove hci{va2_sriov}]$ 
```

fultonj: I've tested in my environment and I see that `kustomize build architecture/examples/va/hci` produces effectively the same yaml in `va2_sriov` branch as the `main` branch. So the change to the HCI VA1 here is fine by me.

fultonj (commented):
Could `architecture/README.md` be updated to link to `architecture/examples/va/nfv/sriov` under "The following VAs are available"?

examples/va/nfv/sriov/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+3. In stages 1 and 2 [kustomize](https://kustomize.io/) is used to genereate the CRs dynamically. The `*-values.yaml` file(s) must be updated to fit your environment. kustomize version 5 or newer required.
+
+4. In stage 3 YAML comments are placed throughout the CRs to aid in the process of customizing the CRs.  Fields that _must_ (or most likely need to be) changed are commented with "# CHANGEME" either on the field itself or somewhere nearby.  Other comments are added to explain fields that can be changed and, sometimes, to explain additions that can be made.
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install the OpenStack K8S operators and their dependencies](../../common/)
```
fultonj, line 28: `../../../common`

Because it's nfv/sriov there's an additional directory to path up to.

examples/va/nfv/sriov/control-plane.md
fultonj: > Perhaps the common directory should have a markdown file describing individual workarounds and both this readme and the HCI read me could link to it. That would avoid content duplication.

Or in `docs` but I don't want to hold this review up over that.

examples/va/hci/kustomization.yaml
abays: I will see what I can do to eliminate that

abays: Fixed

examples/va/nfv/sriov/control-plane.md
abays: Fixed

examples/va/nfv/sriov/README.md
abays: Fixed

examples/va/nfv/sriov/control-plane.md
abays: Yes, let's clean that up in a follow-up

examples/va/nfv/sriov/kustomization.yaml
abays: Fixed

lib/networking/kustomization.yaml
abays: Correct.  My thinking here was that the base `lib` templates should only have common elements that all DTs/VAs would require.  Thus I removed `StorageMgmt` stuff and explicitly added it to the HCI `kustomization.yaml`.  We can discuss this further if we think there is perhaps a better approach (or if you think the old way was preferable).

examples/va/nfv/sriov/network-values.yaml
abays: Yes, I think we could use `general-values` or maybe `infra-values` or something.  Or perhaps we could just remove this one non-network value (`storageClass`) and place it in a new `storage-values`?  I guess my concern is that the direction in which I've proposed to move things here might become unwieldy, i.e. we fragment `values.yaml` into too many separate files.

fultonj: Could we just go back to using `values.yaml`? We can let `service-values.yaml` be it's own thing.

examples/va/nfv/sriov/service-values.yaml
```diff
@@ -0,0 +1,20 @@
...
+    customServiceConfig: |
+      [ml2]
+      mechanism_drivers = ovn,sriovnicswitch
+      [ml2_type_vlan]
+      network_vlan_ranges = sriov-phy4
+  nova:
+    scheduler:
+      customServiceConfig: |
+        [filter_scheduler]
+        enabled_filters = AvailabilityZoneFilter, ComputeFilter, ComputeCapabilitiesFilter, ImagePropertiesFilter, ServerGroupAntiAffinityFilter, ServerGroupAffinityFilter, PciPassthroughFilter, AggregateInstanceExtraSpecsFilter
```
SeanMooney, line 19: so this shoudl be remvoed

https://github.com/openstack-k8s-operators/nova-operator/blob/main/templates/nova.conf#L78

we are configurting the following by default

[filter_scheduler]
available_filters = nova.scheduler.filters.all_filters
enabled_filters = ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,ServerGroupAntiAffinityFilter,ServerGroupAffinityFilter,SameHostFilter,DifferentHostFilter,PciPassthroughFilter,NUMATopologyFilter
# until we can disabel upcalls we can't turn this off by default
# track_instance_changes = false
shuffle_best_same_weighed_hosts = true

the AvailabilityZoneFilter is deprecated and removed form bobcat.
while it is available in antelope/18 we intentionally do not enable it you are also missing the NumaToplogyFilter

while not strictly required its highly likely that sriov will be combined with CPU pinning, hugepages or VMs with a numa topology.

enabling any of those feature without the NumaToplogyFilter is incorrect.

we made sure our details in the nova operator were correct for almost all uses out of the box 

as a general rule i would avoid changing any config option in nova unless expressly need for the va such as scheduled filters.

in this case our default are more correct.

examples/va/nfv/sriov/edpm/nova_sriov.yaml
```diff
@@ -0,0 +1,43 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cpu-pinning-nova
+data:
+  25-cpu-pinning-nova.conf: |
+    [compute]
+    cpu_shared_set = 0-3,24-27 # CHANGEME
+    cpu_dedicated_set = 8-23,32-47 # CHANGEME
+    reserved_host_memory_mb = 4096
```
SeanMooney, line 20: this is not the corect way to reserve host memory on a numa host

and since you have enabled CPU pinning you also need the numatoplogy filter or the scheduler will select host that cant actually fit the VM
due to memory/sriov device affintiy

examples/va/nfv/sriov/edpm/nova_sriov.yaml
```diff
@@ -0,0 +1,43 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sriov-nova
+data:
+  03-sriov-nova.conf: |
+    [pci]
+    passthrough_whitelist = { "devname":"eno4", "physical_network":"sriov-phy4", "trusted":"true" } # CHANGEME
```
SeanMooney, line 29: "devname":"eno4", is deprecated in 17.1 downstream and it will be unsupported in 18 just an FYI so we should not use devname or passthough_whitelist in the VA
```
[PCI]
passthrough_whitelist
```
and 
```
[DEFAULT]
pci_passthrough_whitelist
```
are both replaced by 
```
[pci]
device_spec
```

i.e.
```
device_spec = {"vendor_id":"a2d6",
                   "product_id":"15b3",
                   "address": "0000:82:00.0",
                   "physical_network":"physnet1",
               }
```

SeanMooney:
im confilcited between if we should revert this and then fix it properly before its merged
or just fix the issues in a followup pr.

as documented this VA would not be supported in our downstream product as it is deploying nova in a way that is no longer supported.

PR #64, opened 2024-01-11
Use vanity names for L2 advertisement

leifmadsen:
Use the interface vanity names referenced by NNCP within the L2 advertisement configuration.

### DIFF
--- a/examples/va/hci/values.yaml
+++ b/examples/va/hci/values.yaml
@@ -71,7 +71,7 @@ data:
         vlan: 20
     mtu: 1500
     prefix-length: 24
-    iface: enp7s0.20
+    iface: internalapi
     vlan: 20
     base_iface: enp7s0
     lb_addresses:
@@ -104,7 +104,7 @@ data:
         vlan: 21
     mtu: 9000
     prefix-length: 24
-    iface: enp7s0.21
+    iface: storage
     vlan: 21
     base_iface: enp7s0
     lb_addresses:
@@ -143,7 +143,7 @@ data:
         vlan: 22
     mtu: 1500
     prefix-length: 24
-    iface: enp7s0.22
+    iface: tenant
     vlan: 22
     base_iface: enp7s0
     lb_addresses:


abays:
This fixes a problem caused by a combination of hardcoded vanity interface names in the `NNCP` template [1] and the use of the `values.yaml` `data.(internalapi|storage|tenant).iface` in the `L2Advertisement` template for `spec.interfaces.0` [2].  The `L2Advertisement` interface names need to match the vanity interface names created by the `NNCP`.

[1] https://github.com/openstack-k8s-operators/architecture/blob/d1bd4f70ded050463064d929c7342ccbcb660bff/lib/nncp/ocp_node_template.yaml#L26
[2] https://github.com/openstack-k8s-operators/architecture/blob/d1bd4f70ded050463064d929c7342ccbcb660bff/lib/networking/kustomization.yaml#L174-L184

abays:
Another option is to modify the `NNCP`'s `kustomizational.yaml` to have it consume `data.(internalapi|storage|tenant).iface` [1].  Then the user's choice of interface name is applied globally and we remove the problem caused by the hardcoded interface name in the `NNCP` template.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/nncp/kustomization.yaml

fultonj:
+2 makes sense to me.

fultonj:
> +2 makes sense to me.

This patch should however update both `values.yaml` and `values-ci-framework.yaml`.

fultonj (changes requested):
Please apply the same change in `values.yaml` to `values-ci-framework.yaml`

fultonj:
> Please apply the same change in `values.yaml` to `values-ci-framework.yaml`

Apologies,  `values-ci-framework.yaml` already has this change.

I'm OK to merge as-is.

fultonj:
> Another option is to modify the `NNCP`'s `kustomizational.yaml` to have it consume `data.(internalapi|storage|tenant).iface` [1]. Then the user's choice of interface name is applied globally and we remove the problem caused by the hardcoded interface name in the `NNCP` template.
> 
> [1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/nncp/kustomization.yaml

I propose we merge this patch as is. We could have a follow up though to do what's described above.

PR #45, opened 2023-12-14
Uni* DTs: a bunch of proven, multicomponent set of deployments

tosky:
The deployment part of this proposal have been used to combine and test together a good amount of parts which "makes sense" together.
In fact what is available now as VA1 was originally inspired by one of these topologies (if it wouldn't have been created as VA1, that would have been unigamma).

PR #65, opened 2024-01-11
Multibackend stores DT

maximsava12:
This is draft PR for creating DT

### DIFF
--- a/examples/dt/glance/multibackend/multibackend-dt.md
+++ b/examples/dt/glance/multibackend/multibackend-dt.md
@@ -1,0 +1,60 @@
+## General information
+
+| Revision | Change                |    Date    |
+|--------: | :-------------------- |:----------:|
+| v0.1     | Initial publication   | 2024-01-11 |
+
+## Purpose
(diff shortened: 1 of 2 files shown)


tosky:
Couldn't this be handled by one of the DTs proposed in https://github.com/openstack-k8s-operators/architecture/pull/45, for example extending uniepsilon?

Also, do tempest tests for glance properly handle different glance backends? Or do they just use a random backend?

tosky:
I see a change to VA1, is that intended? Also, is this still needed or could it be merged into another DT as proposed?

maximsava12:
Yes,the change is intended.Not sure why i see the fix network PR here )). 
The can merge to another DT

maximsava12:
Close it, as this DT implemented in  Uni DT epsilon.
https://github.com/openstack-k8s-operators/architecture/pull/45/commits/184193a7c2c91c6a9a39cd8043c9ad1591ee8432#diff-1abbe2ddd94883298848665c0dfa4994d6509025bfc3441a8d89d1612b0aa065

PR #71, opened 2024-01-16
Move required VA-specific kustomizations out of examples

abays:
Moves certain required customizations to the underlying `lib` templates from the `examples/va/*` dirs to the root `va` dirs themselves.  This is because for HCI and SRIOV, the customizations in question are not optional and are thus properly located in the specific `va` subdir for the respective VA.  What is contained in the `examples` dir should be additional customizations and, of course, the various `values` YAMLs.  This allows for the following templating hierarchy:

1. `lib` (base templates common to all VAs and DTs)
2. `va/*` (VA-specific templates per unique VA) or `dt/*` (DT-specific templates per unique DT)
3. `examples/va/*` or `examples/dt/*` (environment-specific templates per unique user environment)


fultonj (approved):
Thank you @abays

PR #72, opened 2024-01-16
Update main README overview

leifmadsen:
Update the main README overview with some additional verbosity and
clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,30 @@
 # OpenStack K8S Operators Architectures
 
-This repository may be used with to create validated
-architectures ([VAs](examples/va)), represented as CRs for
-[openstack-k8s-operators](https://github.com/openstack-k8s-operators).
-It may also be used to create deployed topologies
-([DTs](examples/dt)) which should only be used for testing.
+This repository contains templates used a a part of the validated architecture
+(VA) and deployment topology (DT) effort.
+
+Validated architectures and deployment topologies are represented as kustomize
+compatible templates and custom resources.
+
+Deployment topologies are only to be used for testing environments. Validated
+architectures are intended to represent production-style deployment
+environments.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


leifmadsen:
Actually @abays has a really good overview to help understand the repository structure, so I'm going to add that here as well.

```
1. Base templates in lib that are common to all VAs and DTs
2. VA and DT specific templates in va/* and dt/* that are specific to a given VA or DT
3. User-environment templates and values in examples/va/* and examples/dt/* that are specific to a given VA or DT as it would be deployed in the user's unique environment
```

Let me hack at this a bit and add a new section, as I feel it'd be useful to understand the directory structure.

README.md
```diff
@@ -1,14 +1,30 @@
 # OpenStack K8S Operators Architectures
 
-This repository may be used with to create validated
-architectures ([VAs](examples/va)), represented as CRs for
-[openstack-k8s-operators](https://github.com/openstack-k8s-operators).
-It may also be used to create deployed topologies
-([DTs](examples/dt)) which should only be used for testing.
+This repository contains templates used are part of the validated architecture
```
fultonj, line 3: "used as part of the"  instead of "used are part of"

fultonj (commented):
This is good, but please make one grammar change first.

README.md
leifmadsen, line 3:
```suggestion
This repository contains templates used as a part of the validated architecture
```

leifmadsen: @fultonj so this is how I normally handle suggestions in case you're using the UI. I'm not sure how picky we are about flat commit logs. In STF we prefer verbosity and history over commit cleanliness, but that might not be the expectation here. My default is using the suggestion method (which provides an extra commit) and then squashing commits at the end, if there isn't already a strong preference.

PR #73, opened 2024-01-17
Neutron DT

bharathmv13:
This PR is to add the Neutron DTs

### DIFF
--- a/examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
+++ b/examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
@@ -1,0 +1,67 @@
+# Deployed Topology 3cont_2comp-ipv6-all-geneve-dvr-stf-rally
+
+## General information
+
+| Revision | Change              |  Date      |
+| -------: | :------------------ | :-----:    |
+|     v0.1 | Initial publication | 2024-01-16 |
+|     v0.2 | Refreshed doc       | 2024-02-29 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN                      | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+### Additional information 
+tenant_network = geneve
+public_network = flat
+Enable Service Telemetry
+Add rally tests when available
+This DT is mapped to unified job - uni-delta-ipv6
+
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+add configuration of ipv6, OVNEncapTos=inherit
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Rally            | stage7       |                       |      11111111       | 
+| Tempest/neutron  | stage6       | Use rhel image        |      11223344       |
+| Tempest/scenario | stage6       | Use cirros image      |      22334455       |
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Configure and deploy the control plane](stage4)
+5. [Post Tasks] Create necessary resources for testing (stage5)
+6. [Execute non destructive testing](stage6)
+7. [Execute load testing](stage7)
+8. [Destructive testing](stage8)
--- a/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-Naz
+++ b/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-Naz
@@ -1,0 +1,69 @@
+# Deployed Topology 3cont_2comp_3net-ipv4-geneve-composable-NAz
+
+## General information
+
+| Revision | Change              |  Date      |
+| -------: | :------------------ | :-----:    |
+|     v0.1 | Initial publication | 2024-01-16 |
+|     v0.2 | Refreshed doc       | 2024-02-29 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+| Networker                                  | vm    | 3      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN, neutron_azs_ovn     | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+
+### Additional information 
+tenant_network = geneve
+public_network = flat
+This DT has 3 networker nodes, hence ovn-controller is disabled in controller node
+This DT is mapped to unified job - uni-alpha
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+OVNEmitNeedToFrag = True
+ovn_emit_need_to_frag = True
+[The above specs are taken from network-environment_v2.yaml]
+
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage6       | Use rhel image        |      11223344       |
+| Tempest/scenario | stage6       | Use cirros image      |      22334455       |
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Configure and deploy the control plane](stage4)
+5. [Post Tasks] Create necessary resources for testing (stage 5)
+6. [Execute non destructive testing](stage6)
+7. [Execute load testing](stage7)
+8. [Destructive testing](stage8)
--- a/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network
+++ b/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network
@@ -1,0 +1,70 @@
+# Deployed Topology 3cont_2comp_3net-ipv4-geneve-composable-vlan-provider-network
+
+## General information
+
+| Revision | Change              |  Date      |
+| -------: | :------------------ | :-----:    |
+|     v0.1 | Initial publication | 2024-01-16 |
+|     v0.2 | Refreshed doc       | 2024-02-29 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+| Networker                                  | vm    | 3      |
+| CFME-tester                                | vm    | 1      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN                      | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+
+### Additional information 
+tenant_network = geneve
+public_network = vlan
+This DT has 3 networker nodes, hence ovn-controller is disabled in controller node
+This DT needs a CFME-tester node to act as a router
+This DT is mapped to unified job - uni-eta
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+OVNEmitNeedToFrag = True
+ovn_emit_need_to_frag = True
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage7       | Use rhel image        |      11223344       |
+| Tempest/scenario | stage7       | Use cirros image      |      22334455       |
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Router Network Config] Configure CFME-tester as router (stage4)
+5. [Configure and deploy the control plane](stage5)
+6. [Post Tasks] Create necessary resources for testing (stage6)
+7. [Execute non destructive testing](stage7)
+8. [Execute load testing](stage8)
+9. [Destructive testing](stage9)
--- a/examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
+++ b/examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
@@ -1,0 +1,85 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  neutron:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        service_plugins = qos,ovn-router,trunk,segments,port_forwarding,log
+        dhcp_agent_notification = false
+        agent_down_time = 600
+        router_distributed = true
+        router_scheduler_driver = neutron.scheduler.l3_agent_scheduler.ChanceScheduler
+        allow_automatic_l3agent_failover = true
+        debug = true
+        log_dir = /var/log/neutron
+        default_availability_zones = zone-1,zone-2
+
+        [agent]
+        report_interval = 300
+
+        [database]
+        max_retries = -1
+        db_max_retries = -1
+
+        [keystone_authtoken]
+        region_name = regionOne
+        memcache_use_advanced_pool = True
+        auth_type = password
+
+        [nova]
+        region_name = regionOne
+        endpoint_type = internal
+
+        [oslo_messaging_notifications]
+        driver = noop
+
+        [oslo_middleware]
+        enable_proxy_headers_parsing = true
+
+        [oslo_policy]
+        policy_file = /etc/neutron/policy.yaml
+
+        [placement]
+        region_name = regionOne
+
+        [ovs]
+        igmp_snooping_enable = true
+
+        [ovn]
+        ovsdb_probe_interval = 60000
+        ovn_emit_need_to_frag = true
+        ovn_metadata_enabled = true
+        enable_distributed_floating_ip = true
+
+        [ml2]
+        type_drivers = geneve,vxlan,vlan,flat
+        tenant_network_types = geneve,flat
+        extension_drivers = qos,port_security,dns_domain_ports
+
+        [ml2_type_flat]
+        flat_networks = datacentre
+
+        [ml2_type_geneve]
+        vni_ranges = 1:65536
+        max_header_size = 38
+
+        [ml2_type_vlan]
+        network_vlan_ranges = tenant:1000:2000
+
+        [ml2_type_vxlan]
+        vni_ranges = 1:65536
+        vxlan_group = 224.0.0.1
+
+  nova:
+    template:
+      schedulerServiceTemplate:
+        customServiceConfig: |
+          [filter_scheduler]
+          enabled_filters = AggregateInstanceExtraSpecsFilter, AvailabilityZoneFilter,ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,DifferentHostFilter,SameHostFilter,AllHostsFilter,ServerGroupAffinityFilter,ServerGroupAntiAffinityFilter
+
+
--- a/examples/dt/neutron/CRD_files/neutron_ovn.yaml
+++ b/examples/dt/neutron/CRD_files/neutron_ovn.yaml
@@ -1,0 +1,79 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+    spec:
+      neutron:
+        template:
+        customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        service_plugins = qos,ovn-router,trunk,segments,port_forwarding,log
+        dhcp_agent_notification = false
+        agent_down_time = 600
+        router_distributed = true
+        router_scheduler_driver = neutron.scheduler.l3_agent_scheduler.ChanceScheduler
+        allow_automatic_l3agent_failover = true
+        debug = true
+        log_dir = /var/log/neutron
+
+        [agent]
+        report_interval = 300
+
+        [database]
+        max_retries = -1
+        db_max_retries = -1
+
+        [keystone_authtoken]
+        region_name = regionOne
+        memcache_use_advanced_pool = True
+        auth_type = password
+
+        [nova]
+        region_name = regionOne
+        endpoint_type = internal
+
+        [oslo_messaging_notifications]
+        driver = noop
+
+        [oslo_middleware]
+        enable_proxy_headers_parsing = true
+
+        [oslo_policy]
+        policy_file = /etc/neutron/policy.yaml
+
+        [placement]
+        region_name = regionOne
+
+        [ovs]
+        igmp_snooping_enable = true
+
+        [ml2]
+        type_drivers = geneve,vxlan,vlan,flat
+        tenant_network_types = vlan,flat
+        extension_drivers = qos,port_security,dns_domain_ports
+
+        [ml2_type_flat]
+        flat_networks = datacentre
+
+        [ml2_type_geneve]
+        vni_ranges = 1:65536
+        max_header_size = 38
+
+        [ml2_type_vlan]
+        network_vlan_ranges = tenant:1000:2000
+
+        [ml2_type_vxlan]
+        vni_ranges = 1:65536
+        vxlan_group = 224.0.0.1
+
+
+  nova:
+    template:
+      schedulerServiceTemplate:
+        customServiceConfig: |
+          [filter_scheduler]
+          enabled_filters = AggregateInstanceExtraSpecsFilter, AvailabilityZoneFilter,ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,DifferentHostFilter,SameHostFilter,AllHostsFilter,ServerGroupAffinityFilter,ServerGroupAntiAffinityFilter
+
+
--- a/examples/dt/neutron/3cont_2comp-ipv4-vlan-dvr
+++ b/examples/dt/neutron/3cont_2comp-ipv4-vlan-dvr
@@ -1,0 +1,63 @@
+# Deployed Topology 3cont_2comp-ipv4-vlan-dvr-ovn
+
+## General information
+
+| Revision | Change              |  Date      |
+| -------: | :------------------ | :-----:    |
+|     v0.1 | Initial publication | 2024-01-10 |
+|     v0.2 | Refreshed doc       | 2024-02-29 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN                      | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+### Additional information 
+tenant_network = vlan
+public_network = flat
+This DT is mapped to unified job - uni-delta
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+add configuration to enable compute node to access external network via br-ex
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage6       | Use rhel image        |      11223344       |
+| Tempest/scenario | stage6       | Use cirros image      |      22334455       |
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Configure and deploy the control plane](stage4)
+5. [Post Tasks] Create necessary resources for testing (stage5)
+6. [Execute non destructive testing](stage6)
+7. [Execute load testing](stage7)
+8. [Destructive testing](stage8)
--- a/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network-no_dvr
+++ b/examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network-no_dvr
@@ -1,0 +1,71 @@
+# Deployed Topology 3cont_2comp_3net-ipv4-geneve-composable-vlan-provider-network_no-dvr
+
+## General information
+
+| Revision | Change              |  Date      |
+| -------: | :------------------ | :-----:    |
+|     v0.1 | Initial publication | 2024-04-01 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+| Networker                                  | vm    | 3      |
+| CFME-tester                                | vm    | 1      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN,no-dvr               | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+
+### Additional information 
+tenant_network = geneve
+public_network = vlan
+router_distributed = False
+This DT has 3 networker nodes, hence ovn-controller is disabled in controller node
+This DT needs a CFME-tester node to act as a router
+This DT is mapped to unified job - uni-zeta
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+OVNEmitNeedToFrag = True
+ovn_emit_need_to_frag = True
+enable_distributed_floating_ip = False
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage7       | Use rhel image        |      11223344       |
+| Tempest/scenario | stage7       | Use cirros image      |      22334455       |
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Router Network Config] Configure CFME-tester as router (stage4)
+5. [Configure and deploy the control plane](stage5)
+6. [Post Tasks] Create necessary resources for testing (stage6)
+7. [Execute non destructive testing](stage7)
+8. [Execute load testing](stage8)
+9. [Destructive testing](stage9)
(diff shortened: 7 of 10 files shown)


gurprsin (approved):

tosky:
Before approving any DTs, shouldn't we discuss for opportunities on combining it with other DTs?
In this case, I think the scope of the DTs proposed here is not always fully clear.

booxter:
I have perhaps simple questions, but I have to understand this first before I can meaningfully review this DT, or any DT for that matter.

1. Reading README for this repo, and looking at examples already in the repo and proposed as open PRs, I conclude that the main goal / meat of a DT is a set of k8s yaml templates that can be applied against a OCP cluster to get it to a desired state-under-test. Yes, these templates may have markdown files describing the DTs attached, but it cannot be the core of a DT. (I see a single markdown file included here. Though without .md extension. Is it all there is to have for a new DT, or is the PR incomplete?)

2. I would like to understand the methodology of how the DTs are proposed in this PR. I understand that the proposal is somehow based on existing CI jobs for previous versions of RH-OSP. On the other hand, it seemed to me that the intent is to critically revisit the set of existing jobs, come up with the list of desired tests that we would like to migrate to RHOSO, then express the result in yaml format here as a new DT (or incorporate the result into existing DTs). I also assumed that the list of desired tests would be built based on a list of desired (supported) features in RHOSO. Is my thinking correct? Do we have the list of desired (supported) features that we can cross-check against to confirm the migrated jobs are desired?

If you have some materials for me to read on about the intended methodology for new DTs, I would love to see these. Or perhaps you could explain how you decided what should be part of this PR? Is there some documented guidance you used to produce this PR? What I am missing to proceed with a more substantial review is both the why and the how. (Why = intent of the PR, what is the end goal of these new files; and how = the logical steps that lead to producing these DTs, and not some other DTs.) I hope you can enlighten me.

bharathmv13:
Hi Ihar,

Let me give you my comments on the way i created these DTs:

1. Consider that there are no new RFE implemented in OSP 18.0 that are not there in 17.1
2. Consider that there are no OVS jobs in OSP 18
3. We start with 7 Virtual based jobs and 1 hybrid job(SRIOV)
4. Removing the jobs which are already consolidated, gives us 5 virtual jobs.
5. In that 1 job is vxlan, which no customer is using as per mail threads. so we end up with 4 topologies

rhel-virthost-3cont_2comp-ipv4-vxlan-ovn
rhel-virthost-3cont_2comp-ipv6-all-geneve-dvr-stf-rally
rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls-dvr-ovn
rhel-virthost-3cont_2comp_3net-ipv4-geneve-composable-Naz


These jobs have been translated to DTs in the PR submitted by me.

Please let me know if you need any more info on this.

> I have perhaps simple questions, but I have to understand this first before I can meaningfully review this DT, or any DT for that matter.
> 
> 1. Reading README for this repo, and looking at examples already in the repo and proposed as open PRs, I conclude that the main goal / meat of a DT is a set of k8s yaml templates that can be applied against a OCP cluster to get it to a desired state-under-test. Yes, these templates may have markdown files describing the DTs attached, but it cannot be the core of a DT. (I see a single markdown file included here, per "DT". Though without .md extension. Is it all there is to have for a new DT, or is the PR incomplete?)
> 2. I would like to understand the methodology of how the DTs are proposed in this PR. I understand that the proposal is somehow based on existing CI jobs for previous versions of RH-OSP. On the other hand, it seemed to me that the intent is to critically revisit the set of existing jobs, come up with the list of desired tests that we would like to migrate to RHOSO, then express the result in yaml format here as a new DT (or incorporate the result into existing DTs). I also assumed that the list of desired tests would be built based on a list of desired (supported) features in RHOSO. Is my thinking correct? Do we have the list of desired (supported) features that we can cross-check against to confirm the migrated jobs are desired?
> 
> If you have some materials for me to read on about the intended methodology for new DTs, I would love to see these. Or perhaps you could explain how you decided what should be part of this PR? Is there some documented guidance you used to produce this PR? What I am missing to proceed with a more substantial review is both the why and the how. (Why = intent of the PR, what is the end goal of these new files; and how = the logical steps that lead to producing these DTs, and not some other DTs.) I hope you can enlighten me.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+|     v0.1 | Initial publication | 2024-01-16 |
+
+## Node topology
+
+| Node role                                  | bm/vm | amount |
+| ------------------------------------------ | ----- | ------ |
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+| Networker                                  | vm    | 3      |
+| CFME tester                                | vm    | 1      |
```
rsafrono, line 16: IMO, configuration of CFME tester node (which is actually a router node supporting public network of type 'vlan') can be a  separate stage in the list of stages below. At least with jenkins+infrared we had a separate stage for configuring an external router. Probably with moving to OCP infrastructure we can just configure networks accordingly and get rid of this node, but not 100% sure.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN, VLAN                | Must have          |
+|           | performance                  |                    |
+| Glance    | default                      | Must have          |
+| Cinder    | cinder_backup                | Must have          |
```
rsafrono, line 26: I am not sure that this job actually requires this setting https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/vars/overcloud/templates/cinder_backup.yml
IMO, it was just copy-pasted, but better to check with @eolivare

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+| Networker                                  | vm    | 3      |
+| CFME tester                                | vm    | 1      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN, VLAN                | Must have          |
+|           | performance                  |                    |
```
rsafrono, line 24: Performace tweaks for different services, uses these tripleo parameters https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/vars/overcloud/templates/performance.yml
Not sure if the job really requires them, but better to check with @eolivare, since it was added to d/s job config by his patch.
IMO, it would be enough if some of consolidated jobs will use the settings.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage7       | Use rhel image        |      11223344       |
+| Tempest/scenairo | stage7       | Use cirros image      |      22334455       |
```
rsafrono, line 52: a copy-pasted typo in all files, should be 'scenario'

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+| Glance    | default                      | Must have          |
+| Cinder    | cinder_backup                | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+|           | neutron_azs_ovn              |                    |
+
+
+### Additional information 
+composable_roles
```
rsafrono, line 34: composable_roles is a tripleo term, we are using it in this job mainly for testing neutron with standalone networker nodes

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
```diff
@@ -0,0 +1,66 @@
...
+
+## Considerations/Constraints
+
+
+## Testing tree
+Tempest and Rally
+
+| Test framework   | Stage to run | Special configuration | Test case to report |
+| ---------------- | ------------ | --------------------- | :-----------------: |
+| Tempest/neutron  | stage7       | Use rhel image        |      11223344       |
```
rsafrono, line 51: A note for reviewers from other DFGs, all neutron d/s jobs use cirros as main(default) tempest image, rhel image is configured as an advanced image that is used only for tests where it is required. Tests that use  cirros and rhel are mixed in the same results file.
The above is related to all neutron jobs except the one that runs Rally, that one uses rhel as main tempest image. This was made because rally deploys some specific cirros image and there was some conflict during tempest deploy.

examples/dt/neutron/3cont_2comp_1ipa-ipv4-vlan-dvr
```diff
@@ -0,0 +1,63 @@
...
+### Additional information 
+This DT uses FreeIPA
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+add configuration of ipv4
+--network-force-compute-dvr yes
```
rsafrono, line 37: IMO, we can omit this since the job has dvr enabled. Formerly (before 17.1) this job was with centralized routers and we nevertheless wanted compute nodes to be connected via br-ex to external network. Now this is a redundant setting.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
tosky: This is good to read because it means we can combine these DTs with others. About rally, wasn't it possible to patch it to fix the conflict?

examples/dt/neutron/3cont_2comp_1ipa-ipv4-vlan-dvr
```diff
@@ -0,0 +1,63 @@
...
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN, VLAN                | Must have          |
+| Glance    | default                      | Must have          |
+| Cinder    | default                      | Must have          |
+| Keystone  | default                      | Must have          |
+| TLS       | default                      | Must have          |
+| Compute   | DifferentHostFilter          | Must have          |
+
+### Additional information 
+This DT uses FreeIPA
```
rsafrono, line 29: why freeipa? IIRC, TLS is enabled in all osp18 jobs by default so IIUC no specific config is needed from our side for this.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
tosky: if it's not needed it would indeed be better to leave it out, so that it's easier to consolidate this with other DTs.

tosky:
I see

    In that 1 job is vxlan, which no customer is using [...] so we end up with 4 topologies

but then you list rhel-virthost-3cont_2comp-ipv4-vxlan-ovn, which looks like ovn/vxlan?

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
```diff
@@ -0,0 +1,62 @@
...
+| Compute   | DifferentHostFilter          | Must have          |
+
+### Additional information 
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+add configuration of ipv6, OVNEncapTos=inherit
```
rsafrono, line 36: IMO, no specific configuration we need to add to OVN for ipv6, as well as ipv4 mentioned for ovn in other jobs. Since ipv4/ipv6 here is the control plane setting it should be pretty transparent for neutron. I think we should just inherit some global ipv6 control plane config  (or ipv4 for other jobs)  that will be common for most jobs.

rsafrono:
> I see
> 
> ```
> In that 1 job is vxlan, which no customer is using [...] so we end up with 4 topologies
> ```
> 
> but then you list rhel-virthost-3cont_2comp-ipv4-vxlan-ovn, which looks like ovn/vxlan?

@tosky the list is wrong, it contains ovn/vxlan and does not contain vlan-provider job. Actual files in commit are correct, I mean there is no DT for ovn/vxlan and a DT for vlan-provider is available.

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-Naz
```diff
@@ -0,0 +1,64 @@
+# Deployed Topology 3cont_2comp_3net-ipv4-geneve-composable-Naz
```
booxter, line 1: what is "Naz"?

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
```diff
@@ -0,0 +1,62 @@
...
+| Openshift master/worker combo-node cluster | vm    | 3      |
+| Compute nodes                              | vm    | 2      |
+
+
+## Services, enabled features and configurations
+
+| Service   | configuration                | Lock-in coverage?  |
+| --------- | ---------------------------- | ------------------ |
+| Neutron   | ML2/OVN, Geneve              | Must have          |
+|           | ovn-router-flavors           |                    |
```
booxter, line 22: what's the reason why router-flavors are enabled in this job but not in others? is there some test scenario that fails with it enabled?

booxter:
> rhel-virthost-3cont_2comp-ipv4-vxlan-ovn
> rhel-virthost-3cont_2comp-ipv6-all-geneve-dvr-stf-rally
> rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls-dvr-ovn
> rhel-virthost-3cont_2comp_3net-ipv4-geneve-composable-Naz

(I don't think the list matches what is proposed here, but I will attempt to interpret it)

If I understand correctly the intent, you have

- a job with geneve tenant networks
- a job with vlan tenant networks

for each of:

- setup with no networker nodes
- setup with 3 dedicated EDP networker nodes.

Which gives us 2x2 = 4 total jobs. Is it the only DESIGNED difference between these jobs? (I mean - are the rest of changes, e.g. IPA or introduction of a separate machine to configure upstream router ACCIDENTAL, even if perhaps needed to make it all work with the code and bugs present?) Is there something that I missed?

If this is the test matrix we are targeting, (2x2 on the axis of "vlan vs geneve" and "EDP networker vs no networkers"), then this makes a lot of sense. Should we document the intent in the files proposed? (It's not as easy to try to deduce the intent from file names and tables listing needed VMs and configuration, esp. since names are abbreviated and some of them don't reflect the intent anyway.)

Do you plan to add CRs to reflect the proposed DTs descriptions?

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-Naz
rsafrono: IIRC, network (or neutron) availability zones

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
rsafrono: This is aligned with our existing d/s jobs. IIUC, we can have ovn-router-flavors enabled by default in all jobs, but usually we use neutron defaults and enable specific features where needed

rsafrono:
I've made some additional checks and here I am trying to summarize the specifics I found between different jobs:

All of the jobs from the list in osp18 will be using dvr and tls by default so we can omit mentioning 'dvr' and 'tls' 
Also we can enable ovn-router-flavors for all as well as setting that matches OVNEncapTos=inherit tripleo setting
Allof the jobs trigger the same set of tempest tests [infrared suites that we used in 17.1 were neutron, neutron_single_threaded and neutron_disruptive]
Compute service should have DifferentHostFilter enabled in all jobs [various tests require spawning 2 VMs on different hosts for proper testing of some networking features]

3cont_2comp-ipv6-geneve-rally
  tenant network = geneve
  public network = flat
  control plane ip version = ipv6
  runs a suite of rally tests [in addition to usual tempest suites]  [TBD]

3cont_2comp_ipv4-vlan
  tenant network = vlan
  public network = flat
  control plane ip version = ipv4

3cont_2comp_3net-ipv4-geneve-composable-Naz
  tenant network = geneve
  public network = flat
  control plane ip version = ipv4
  3 standalone networkers
  neutron availability zones config, i.e. https://github.com/redhat-openstack/infrared/blob/master/plugins/tripleo-overcloud/vars/overcloud/templates/neutron_azs_ovn.yml
  tenant network mtu 9000, external network mtu 1500
  additional specifics are same as described below for vlan-provider-network job
  
3cont_2comp_3net-ipv4-geneve-composable-vlan-provider-network
  tenant network = geneve
  public network = vlan
  control plane ip version = ipv4
  3 standalone networkers
  tenant network mtu 9000, external network mtu 1500
  additional specifics can be seen at the bottom of this file https://code.engineering.redhat.com/gerrit/gitweb?p=Neutron-QE.git;a=blob;f=vlan_provider_network_ovn/OSP17/network/network-environment_v2.yaml.j2;h=542d070772096d8de59512f630b44b9d2004aa76;hb=refs/heads/master#l40
  i.e. "OVNEmitNeedToFrag: True" and "neutron::plugins::ml2::ovn::ovn_emit_need_to_frag: True"   which are required for d/s tests under neutron_plugin.tests.scenario.test_mtu

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
booxter: indeed, this may be a better question for the product - if we should have it enabled by default.

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
```diff
@@ -0,0 +1,62 @@
...
+
+## Stages
+
+All stages must be executed in the order listed below. Everything is required unless otherwise indicated.
+
+1. [Install dependencies for the OpenStack K8S operators](stage1)
+2. [Install the OpenStack K8S operators](stage2)
+3. [Configuring networking on the OCP nodes](stage3)
+4. [Configure and deploy the control plane](stage4)
+5. [Configure and deploy the initial data plane to prepare for CephHCI installation](stage5)
```
karelyatin, line 58: is ceph really required? or just there for some reusability point? is ceph used to be there in mentioned 17.1 jobs?

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
```diff
@@ -0,0 +1,62 @@
...
+6. [Update the control plane and finish deploying the data plane after CephHCI has been installed](stage6)
+7. [Execute non destructive testing](stage7)
```
karelyatin, line 60: the stage numbers do not match with the table above

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
```diff
@@ -0,0 +1,62 @@
+# Deployed Topology 3cont_2comp-ipv6-all-geneve-dvr-stf-rally
+
```
karelyatin, line 2: Just generic observation, most of the things in these 4 DTs looks common, so seems it would have been better having a base/common DT and other extend it. That way would be quite visible what unnecessary in extended DTs and clubbed into base when found necessary.

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
rsafrono: A correction regarding ipv6. We do configure ovn-encap-ip when using ipv6. It is set to ipv6 address [this only could be achieved in 17.1 by setting infrared tripleo-overcloud parameter --network-protocol ipv6-all]
When using --network-protocol ipv6 in infrared, control-plane was using ipv6 addresses while ovn was still using ipv4 for tunnels. This was causing  test_dscp_inheritance_geneve to fail. 
The OVNEncapTos=inherit setting was also only just for this test. In case the value is not specified, the test is just skipped.

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
tosky: In fact, my suggestion is to look at https://github.com/openstack-k8s-operators/architecture/pull/45 and see how these DTs can be added there. The components which are part of the DTs proposed there don't care much on how networking is configured, as long as it works!

rsafrono: Agree. 
Current defaults are as follows(in addition to tls, dvr):

tenant network = geneve
public network = flat
control plane ip version = ipv4
mtu 1500
no standalone networkers

default neutron config does not have other above-mentioned specifics (e.g. ovn-router-flavors  etc.)
https://github.com/openstack-k8s-operators/neutron-operator/blob/main/templates/neutronapi/config/01-neutron.conf

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
rsafrono: According to @eolivare cinder_backup is not a 'must have' setting here

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
rsafrono: According to @eolivare performance is not a 'must have' setting here

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
m-blue-m: IIRC this feature had enabling and disabling added and defaults to disabled since it had potential wide impact in case there is a bug, we didn't want it to affect many when only a few need these extra customized flavors.
Other jobs check no regressions exist when feature is disabled (most users), while this specific one tests when enabled.

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
bharathmv13: Marking this as Resolved, because this is mapped to unified job uni-delta-ipv6

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
bharathmv13: Marking this as Resolved, because this is mapped to unified job uni-delta-ipv6

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
bharathmv13: Marking this as Resolved, because this is mapped to unified job uni-delta-ipv6 and in that job this config is added.

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
bharathmv13: Marking this as Resolved, because this is mapped to unified job uni-delta-ipv6

examples/dt/neutron/3cont_2comp-ipv6-all-geneve-dvr-stf-rally
bharathmv13: Marking this as Resolved, because this is mapped to unified job uni-delta-ipv6

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
...
+  neutron:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        network_scheduler_driver = neutron.scheduler.dhcp_agent_scheduler.AZAwareWeightScheduler
+        dhcp_load_type = networks
+        default_availability_zones = zone-1,zone-2
+        igmp_snooping_enable = true
```
karelyatin, line 15: Duplicate of line 19, so can be dropped from here as this belong to ovs group

Also i don't see it's mentioned in dt file, from where it's coming?

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: openstack
+spec:
+  neutron:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        network_scheduler_driver = neutron.scheduler.dhcp_agent_scheduler.AZAwareWeightScheduler
```
karelyatin, line 12: this seems specific to dhcp agent

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
```
karelyatin, line 4: I think the content needs to be structured to kustomize uses, but that's someone from maintainers can confirm.

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
...
+        global_physnet_mtu = 9000
+
+        [ovs]
+        igmp_snooping_enable = true
+    
+        [ovn]
+        ovn_emit_need_to_frag = true
+  
+        [ml2]
+        path_mtu = 9000
```
karelyatin, line 25: considering this dt is targetting network nodes, may be we should disable ovn controller on control plane?

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
bharathmv13: This config i found in neutron.conf, however it is disbled, so i will remove this in next commit!

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
bharathmv13: This line will be dropped as it is found only in [ovs] section of neutron.conf.

Should this be mentioned in DT ?

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: openstack
+spec:
+  neutron:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        network_scheduler_driver = neutron.scheduler.dhcp_agent_scheduler.AZAwareWeightScheduler
+        dhcp_load_type = networks
```
slawqo, line 13: this option is also only related to the agents scheduler so not needed here

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
slawqo: I'm not sure where is the information that this DT is targetting deployment with networker nodes but if that's true then yes, we should definitely disable ovn-controller on Podified control plane as networker nodes should do that job :)

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,34 @@
...
+  
+        [ml2]
+        path_mtu = 9000
+
+  nova:
+    template:
+      schedulerServiceTemplate:
+        customServiceConfig: |
+          [filter_scheduler]
+          enabled_filters = RetryFilter, AggregateInstanceExtraSpecsFilter, AvailabilityZoneFilter,ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,DifferentHostFilter,SameHostFilter,AllHostsFilter,ServerGroupAffinityFilter,ServerGroupAntiAffinityFilter
```
rsafrono, line 32: IIRC, RetryFilter is not available starting from osp17, https://bugzilla.redhat.com/show_bug.cgi?id=2028171.

bharathmv13: Got it!, will remove in the next commit! thanks!

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
bharathmv13: This is resolved

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network
```diff
@@ -0,0 +1,72 @@
...
+This DT has 3 networker nodes
+This DT needs a CFME-tester node to act as a router
+This DT is mapped to unified job - uni-eta
+
+#### Nova
+
+#### Cinder
+
+#### OVN
+tenant_network_mtu = 9000
```
rsafrono, line 44: Note, all this enlarged mtu related stuff will require some additional settings in other operators. While initially we planned to keep this setting, now I am thinking that for probably for simplicity we can use a default mtu in unified jobs and just decrease external (aka public or provider) network mtu to some lower value (e.g. 1300). This will fulfill our needs for regression testing of path MTU discovery tests (test_mtu.py in d/s tempest_neutron_plugin plugin or it's corresponding code in whitebox-neutron-tempest-plugin). IMO, we can omit testing jumbo MTU since nfv team covers that.
I.e. I mean maybe we should consider switching here to use tenant_network_mtu = 1442, external_network_mtu = 1300 . The path_mtu and global_physnet_mtu in neutron settings will be default and can be omitted in this case. Wdyt?

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
```diff
@@ -0,0 +1,87 @@
...
+        vlan_transparent = true
+        service_plugins = qos,ovn-router,trunk,segments,port_forwarding,log
+        dhcp_agent_notification = false
+        agent_down_time = 600
+        router_distributed = true
+        router_scheduler_driver = neutron.scheduler.l3_agent_scheduler.ChanceScheduler
+        allow_automatic_l3agent_failover = true
+        debug = true
+        log_dir = /var/log/neutron
+        default_availability_zones = zone-1,zone-2
```
rsafrono, line 20: IIRC,  for this (default_availability_zones = zone-1,zone-2) to work properly it will require also settings in ovs db on compute nodes

booxter: https://github.com/openstack-k8s-operators/edpm-ansible/pull/573

rsafrono:
Since I noticed many duplications, I decided to prepare a summary that will help to distinguish specifics between neutron settings in jobs and probably can help to prepare data suitable for kustomization. Nova settings are the same in all jobs.
I also compared with the current neutron-operator defaults https://github.com/openstack-k8s-operators/neutron-operator/blob/main/templates/neutronapi/config/01-neutron.conf and put comments correspondingly

# COMMON FOR ALL JOBS

[DEFAULT]
dns_domain = openstackgate.local   #(rsafrono) default setting, can be omitted
vlan_transparent = true
dhcp_agent_notification = false    #(rsafrono) default setting, can be omitted
agent_down_time = 600
router_distributed = true
router_scheduler_driver = neutron.scheduler.l3_agent_scheduler.ChanceScheduler
allow_automatic_l3agent_failover = true
debug = true
log_dir = /var/log/neutron

[agent]
report_interval = 300

[database]
max_retries = -1
db_max_retries = -1

[keystone_authtoken]
region_name = regionOne
memcache_use_advanced_pool = True
auth_type = password     #(rsafrono) default setting, can be omitted

[nova]
region_name = regionOne    #(rsafrono) default setting, can be omitted
endpoint_type = internal    #(rsafrono) default setting, can be omitted

[oslo_messaging_notifications]
driver = noop

[oslo_middleware]
enable_proxy_headers_parsing = true

[oslo_policy]
policy_file = /etc/neutron/policy.yaml

[placement]
region_name = regionOne   #(rsafrono) default setting, can be omitted

[ovs]
igmp_snooping_enable = true

[ml2]
extension_drivers = qos,port_security,dns_domain_ports   #(rsafrono) default setting, can be omitted

[ml2_type_geneve]
vni_ranges = 1:65536    #(rsafrono) default setting, can be omitted
max_header_size = 38     #(rsafrono) default setting, can be omitted


# Common for az, vlan_tenant and ipv6 jobs
[ml2_type_flat]
flat_networks = datacentre

[ml2_type_vlan]
network_vlan_ranges = tenant:1000:2000

[ml2_type_vxlan]
vni_ranges = 1:65536         #(rsafrono) default setting, can be omitted
vxlan_group = 224.0.0.1



# common for az, ipv6 and vlan_provider jobs

[DEFAULT]
service_plugins = qos,ovn-router,trunk,segments,port_forwarding,log    #(rsafrono) default setting, can be omitted

[ml2]
type_drivers = geneve,vxlan,vlan,flat


# common for az and ipv6 jobs

[ml2]
tenant_network_types = geneve,flat



# common for jobs with jumbo frames, i.e. AZ and vlan_provider
[DEFAULT]
global_physnet_mtu = 9000   # probably we can omit in case we decide to use normal MTU and just decrease public network MTU for pmtud tests 

[ovn]
ovn_emit_need_to_frag = true   # for testing Path MTU discovery (pmtud)

[ml2]
path_mtu = 9000    # probably we can omit in case we decide to use normal MTU and just decrease public network MTU for pmtud tests


# AZ job only
[DEFAULT]
default_availability_zones = zone-1,zone-2

[ovn]
ovsdb_probe_interval = 60000            # (rsafrono )found only in az job, why?
ovn_metadata_enabled = true             # default setting, can be omitted
enable_distributed_floating_ip = true   # default setting, can be omitted


# VLAN-tenant job only 
[DEFAULT]
service_plugins = qos,ovn-router-flavors-ha,trunk,segments,port_forwarding,log     #(rsafrono) i.e. ovn-router-flavors-ha rather than just ovn-router

[ml2]
tenant_network_types = vlan,flat


# IPv6 job only
[service_providers]
service_provider=L3_ROUTER_NAT:user-defined:neutron.services.ovn_l3.service_providers.user_defined.UserDefined

[ml2]
overlay_ip_version = 6


# VLAN provider job only
[ml2]
type_drivers = geneve,vlan
tenant_network_types = geneve,vlan

[ml2_type_vlan]
network_vlan_ranges = tenant:1000:2000,datacentre:218:218  
 i.e. datacentre:218:218 added comparing to other jobs. Note: this specific vlan 218 was hard-coded in jenkins deployments[1], IIUC was host-only and for ci-frametwork/zuul deployments can be chosen differently according to existing # infrastructure.
 [1] https://code.engineering.redhat.com/gerrit/gitweb?p=rhos-qe-jenkins.git;a=blob;f=jobs/defaults/stages/router_network_config.groovy.inc;h=8fa29c77ee8fb1eca8c417eb265ddb13b9619be6;hb=refs/heads/master#l31

examples/dt/neutron/CRD_files/neutron_ovn.yaml
```diff
@@ -0,0 +1,79 @@
...
+metadata:
+  name: openstack
+    spec:
+      neutron:
+        template:
+        customServiceConfig: |
+        [DEFAULT]
+        dns_domain = openstackgate.local
+        vlan_transparent = true
+        service_plugins = qos,ovn-router-flavors-ha,trunk,segments,port_forwarding,log
```
rsafrono, line 12: IMO these service_plugins settings should be in ipv6 job, i.e the one with service_provider=L3_ROUTER_NAT:user-defined:neutron.services.ovn_l3.service_providers.user_defined.UserDefined
e.g.  ipv6 job http://rhos-ci-logs.lab.eng.tlv2.redhat.com/logs/rcj/DFG-network-neutron-17.1_director-rhel-virthost-3cont_2comp-ipv6-all-geneve-dvr-stf-rally/57/controller-0/var/lib/config-data/puppet-generated/neutron/etc/neutron/neutron.conf.gz

vlan-tenant job does not have these , e.g.

http://rhos-ci-logs.lab.eng.tlv2.redhat.com/logs/rcj/DFG-network-neutron-17.1_director-rhel-virthost-3cont_2comp_1ipa-ipv4-vlan-tls-dvr-ovn/177/controller-0/var/lib/config-data/puppet-generated/neutron/etc/neutron/neutron.conf.gz

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
bharathmv13: Added a comment in DT to disable ovn-controller in controller nodes

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable-vlan_provider_network
bharathmv13: Removed the Jumbo MTU config.

examples/dt/neutron/CRD_files/neutron_ovn.yaml
bharathmv13: Fixed this config

examples/dt/neutron/CRD_files/neutron_az_ovn.yaml
bharathmv13: Unified team will take care of the Kustomize format

examples/dt/neutron/3cont_2comp_1ipa-ipv4-vlan-dvr
bharathmv13: This is fixed

examples/dt/neutron/3cont_2comp_1ipa-ipv4-vlan-dvr
bharathmv13: Fixed

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
bharathmv13: fixed

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
bharathmv13: fixed

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
bharathmv13: fixed

examples/dt/neutron/3cont_2comp_3net-ipv4-geneve-composable_provider_network
bharathmv13: Separate stage added

examples/dt/neutron/CRD_files/neutron_ovn.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+    spec:
+      neutron:
+        template:
+        customServiceConfig: |
```
karelyatin, line 8: some of the config is duplicated with default, and some will not work with the nic-configs, so below is the adjusted one:-
      customServiceConfig: |
        [DEFAULT]
          vlan_transparent = true
          service_plugins = qos,ovn-router,trunk,segments,log
          agent_down_time = 600
          router_distributed = true
          router_scheduler_driver = neutron.scheduler.l3_agent_scheduler.ChanceScheduler
          debug = true
          [agent]
          report_interval = 300
          [database]
          max_retries = -1
          db_max_retries = -1
          [keystone_authtoken]
          region_name = regionOne
          memcache_use_advanced_pool = True
          auth_type = password
          [nova]
          region_name = regionOne
          endpoint_type = internal
          [oslo_messaging_notifications]
          driver = noop
          [oslo_middleware]
          enable_proxy_headers_parsing = true
          [oslo_policy]
          policy_file = /etc/neutron/policy.yaml
          [placement]
          region_name = regionOne
          [ovs]
          igmp_snooping_enable = true
          [ovn]
          ovn_emit_need_to_frag = true
          [ml2]
          type_drivers = geneve,vlan,flat
          tenant_network_types = vlan,flat
          [ml2_type_vlan]
          network_vlan_ranges = datacentre:1000:200

PR #74, opened 2024-01-17
Add reboot-os service to SRIOV VA

abays:
The `reboot-os` service was introduced to the codebase recently and needs to be included in the SRIOV VA in order to properly apply needed kernel args.

### DIFF
--- a/examples/va/nfv/sriov/edpm/openstackdataplanenodeset.yaml
+++ b/examples/va/nfv/sriov/edpm/openstackdataplanenodeset.yaml
@@ -41,6 +41,7 @@ spec:
     - install-os
     - configure-os
     - run-os
+    - reboot-os
     - libvirt
     - ovn
     - neutron-sriov


leifmadsen (commented):
And this is only required in the SRIOV VA and not others?

I think an update to the README might also be in order here so we can track which services are being deployed for the VA, and why.

fultonj:
> And this is only required in the SRIOV VA and not others?

Of the two, yes. HCI doesn't need this.
 
> I think an update to the README might also be in order here so we can track which services are being deployed for the VA, and why.

I think that would be too detailed. We shouldn't have to explain every service which is deployed in this repo. Also, we don't want to maintain two lists which can get out of date with one another.

leifmadsen:
> > I think an update to the README might also be in order here so we can track which services are being deployed for the VA, and why.
> 
> I think that would be too detailed. We shouldn't have to explain every service which is deployed in this repo. Also, we don't want to maintain two lists which can get out of date with one another.

I had only mentioned it because I thought the READMEs had that kind of detail already, but once I looked again it didn't seem to be the case.

PR #76, opened 2024-01-17
Data plane kustomization for SRIOV VA

abays:
A first pass at adding Kustomize for the data plane, demonstrated within the context of the SRIOV VA


lib/data-plane/dataplane-ssh-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+data:
+  authorized_keys: _replaced_
+  ssh-privatekey: _replaced_
+  ssh-publickey: _replaced_
+kind: Secret
+metadata:
+  name: dataplane-ansible-ssh-private-key-secret
```
leifmadsen, line 8: I'm finding the mixed directory/file name of `data-plane` and `dataplane` super awkward...

leifmadsen (commented):
Overall seems to look ok. I don't see any major typos or what looks like invalid references etc.

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -157,8 +94,73 @@ spec:
...
+    migration:
+      ssh_keys:
+        private: |
+          CHANGEME4
+          CHANGEME4
+          CHANGEME4
+        public: |
+          CHANGEME5
+          CHANGEME5
+          CHANGEME5
```
SeanMooney, line 162: we are currently expecting to use a single ssh key for all host for migration 

so we should not have 3 here.

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -94,18 +34,15 @@ spec:
         service_net_map:
           nova_api_network: internal_api
           nova_libvirt_network: internal_api
-        # CHANGEME
         timesync_ntp_servers:
           - hostname: clock.redhat.com
         # CPU pinning settings
-        edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=16 intel_iommu=on iommu=pt isolcpus=4-23,28-47" # CHANGEME
+        edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=16 intel_iommu=on iommu=pt isolcpus=4-23,28-47"
```
SeanMooney, line 40: note that the compute dfg does not support or recommend the use of isolcpus on non realtime systems.

it is also invalid to include any CPUs that are part of the cpu_share_set in isolcpus regardless of if its a real-time system or not

later we are using cpu_shared_set = 0-3,24-27
so in this case they are not overlaping however

since this is not a realtim host i think isolcpus should be removed for this va and only included in an explicit realtime va instead.

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -94,18 +34,15 @@ spec:
...
         edpm_tuned_profile: "cpu-partitioning" 
-        edpm_tuned_isolated_cores: "4-23,28-47" # CHANGEME
+        edpm_tuned_isolated_cores: "4-23,28-47"
```
SeanMooney, line 42: edpm_tuned_isolated_cores in general should include the cores in both
cpu_shared_set and cpu_dedicated_set as well as any core usef for example ovs-dpdk

in this sriov va i think the best approach would be to either not set
edpm_tuned_isolated_cores or set it to 

```
edpm_tuned_isolated_cores: "3-48"
``

File: examples/va/nfv/sriov/edpm/values.yaml [hunk 5]
```diff
@@ -157,8 +94,73 @@ spec:
...
+          CHANGEME4
+          CHANGEME4
+        public: |
+          CHANGEME5
+          CHANGEME5
+          CHANGEME5
+    pci:
+      conf: |
+        [pci]
+        passthrough_whitelist = { "devname":"eno4", "physical_network":"sriov-phy4", "trusted":"true" }
```
SeanMooney, line 166: as noted on https://github.com/openstack-k8s-operators/architecture/pull/62

device name was deprecated in 17.1 downstream and will be unsupported in our downstream product in 18 and 19.
to get the same behaviour you should use the PCI adress of the PF
with the vendor_id and product_id of the VFs.

device_spec = {"vendor_id":"a2d6",
                   "product_id":"15b3",
                   "address": "0000:82:00.0",
                   "physical_network":"physnet1",
               }

this will make all VFs of the PF consumable.

as part of a move to inclusive naming passthrough_whitelist
was replace with device_spec and is no longer supported in 18

we should not use passthrough_whitelist in any future docs or in the va.
while its technically still functional we expect the PCI section to be updated as part of the adoption to use device_sepc instead.

some newer features in osp 19 will only work with device_spec and do not support devname.

we have recommend against using devname for many upstream and downstream release  at this point so hopefully most of our customer have already stopped using device name at this point

SeanMooney (changes requested):

examples/va/nfv/sriov/edpm/values.yaml
abays: Sorry, the format I have chosen here for the `CHANGEME` is confusing.  I used multiple lines merely to draw the eye to the field so that the user knows to change it.  I will switch all of these to use a single `CHANGEME` to avoid giving the impression that multiple keys are expected.

examples/va/nfv/sriov/edpm/values.yaml
abays: Understood.  We will remove the `isolcpus`.

examples/va/nfv/sriov/edpm/values.yaml
abays: Thank you for the update.  Is it still possible to use the current keys in the example, i.e:

```
{ "devname":"eno4", "physical_network":"sriov-phy4", "trusted":"true" }
```

?

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -94,18 +34,15 @@ spec:
         service_net_map:
           nova_api_network: internal_api
           nova_libvirt_network: internal_api
-        # CHANGEME
         timesync_ntp_servers:
           - hostname: clock.redhat.com
         # CPU pinning settings
-        edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=16 intel_iommu=on iommu=pt isolcpus=4-23,28-47" # CHANGEME
+        edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=16 intel_iommu=on iommu=pt isolcpus=4-23,28-47"
         edpm_tuned_profile: "cpu-partitioning"
```
SeanMooney, line 41: kind of a meta comment, with the focus on improving power saving
i am not convinced we should default to cpu-partioning in general
using "cpu-partitioning-powersave" would be a better default for all our VAs
"cpu-partitioning" disables all CPU power management on the host by forcing the performance governor which pins the CPU at 100% frequency

cpu-partitioning-powersave removes explicitly setting the governor and adds 
intel_pstate=passive to the the kernel command line allowing the cpus to idel when not in use.

https://github.com/redhat-performance/tuned/blob/master/profiles/cpu-partitioning-powersave/tuned.conf

it inhrits form cpu-partitioning so its otherwise effectively the same
it was introduced by the nfv dfg at the direct request of some our our larger telcos during the pandemic sue to the rising cost of power and there desire to reduce carbon output as well as cost.

we are planning to ship with https://specs.openstack.org/openstack/nova-specs/specs/2023.1/implemented/libvirt-cpu-state-mgmt.html enabled by default so nova will turn off unused cpu_dedicated_set cores out of the box to save power but ti does not hurt to also use the power saving tuned profile for the rest of the cores that nova will not manage.

examples/va/nfv/sriov/edpm/values.yaml
abays: Okay, I will use the range you suggest as opposed to leaving it out.  That way we are demonstrating the use of the variable.

examples/va/nfv/sriov/edpm/values.yaml
abays: Ah, I see.  Thank you for the further clarification.  We will use `vendor_id`, `product_id` and `address` instead.

examples/va/nfv/sriov/edpm/values.yaml
atyronesmith: @SeanMooney What has changed since 17.1?  Isolcpus has been in use for a long time now for non-realtime NFV deployments.

SeanMooney: nothing we explcity ask that isolcpus not be used on non realtime host for 17.1 as well.
officially the comput dfg do not recommend the use of isolcpu on non-realtime hosts.
we have had several customer issues related too it form telcos misconfiguing it on non nfv systems and nfv systems

examples/va/nfv/sriov/edpm/values.yaml
abays: That's a good idea, but I'm not sure `kustomize` natively supports pulling in a file and injecting it into a template.  It looks like maybe you could write a plugin to do that, though, so perhaps we can add such a thing as a Jira item to follow-up upon: https://stackoverflow.com/questions/68447251/is-there-a-way-to-have-kustomize-merge-the-content-of-a-file-into-a-yaml-file.

lib/data-plane/dataplane-ssh-secret.yaml
fultonj: Yes, I think we should name it dataplane since that's what the [dataplane-operator](https://openstack-k8s-operators.github.io/dataplane-operator) is already using.

fultonj: since we're introducing the `data-plane` with this change, can we change it to `dataplane`?

examples/va/nfv/sriov/edpm/values.yaml
fultonj: Let's inject that (single) key data via the values file for now. If you want to write a KRM function to pick it up via a path then let's do that in a separate PR. This patch doesn't need to block on that however.

Note also for fully automated CI jobs that values.yaml will be provided by the ci-framework.

After this merges I'd like to apply the same pattern for EDPM kustomize to VA1.

lib/data-plane/dataplane-ssh-secret.yaml
abays: Yes, let's use `dataplane` everywhere here

examples/va/nfv/sriov/edpm/values.yaml
SeanMooney: ya this is not a blocker. i tough we were expecting 3 keys which wont work with what currently automated for cold/live migration

inlining the key is less secure but it wont break anything so this is not a blocker we can follow up once the toolign matures more

examples/va/nfv/sriov/edpm/values.yaml
abays: Ack.  Switching to `cpu-partitioning-powersave`.

lib/data-plane/dataplane-ssh-secret.yaml
abays: Done

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -157,8 +94,73 @@ spec:
...
+    - nova-custom-sriov
+    - neutron-sriov
+    - neutron-metadata
+  nova:
+    compute:
+      conf: |
+        [compute]
+        cpu_shared_set = 0-3,24-27
+        cpu_dedicated_set = 8-23,32-47
+        reserved_host_memory_mb = 4096
```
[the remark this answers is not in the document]
abays: Fixed

examples/va/nfv/sriov/edpm/values.yaml
abays: Fixed

examples/va/nfv/sriov/edpm/values.yaml
abays: Fixed

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -158,8 +85,70 @@ spec:
...
+    - nova-custom-sriov
+    - neutron-sriov
+    - neutron-metadata
+  nova:
+    compute:
+      conf: |
+        [DEFAULT]
+        reserved_host_memory_mb = 4096
+        reserved_huge_pages = node:0,size:4,count:4096000
+        reserved_huge_pages = node:1,size:4,count:4096000
```
SeanMooney, line 143: i realised i failed at basic math

orgianly you were reserving 4G total for the host use but the example i provided are now resving 32G for the numa aware part

so i should have done the following
4096MB -> 4193280 KB
419328KB / 4KB pages = 1048320 pages
1048320 pages / 2 numa nodes = 524160 pages per numa node

so if you have 
reserved_host_memory_mb = 4096
and 2 numa nodes on the host you should have
reserved_huge_pages = node:0,size:4,count:524160
reserved_huge_pages = node:1,size:4,count:524160

i should really just modify nova to do this by default so humans don't have to do basic math

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -158,8 +85,70 @@ spec:
...
+        cpu_shared_set = 0-3,24-27
+        cpu_dedicated_set = 8-23,32-47
+    migration:
+      ssh_keys:
+        private: CHANGEME4
+        public: CHANGEME5
+    pci:
+      conf: |
+        [pci]
+        device_spec = {"vendor_id":"8086", "product_id":"1572", "address": "0000:19:00.3", "physical_network":"sriov-phy4", "trusted":"true"}
```
SeanMooney, line 154: +1 on the other chagnes here

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -1,89 +1,19 @@
...
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  ssh_keys:
+    # Authorized keys that will have access to the dataplane computes via SSH
+    authorized: CHANGEME
+    # The private key that will have access to the dataplane computes via SSH
+    private: CHANGEME2
+    # The public key that will have access to the dataplane computes via SSH
+    public: CHANGEME3
```
SeanMooney, line 15: by the way one thing to be aware of is the secretGenerator can use files directly

so if we just want to put these directly into a secreate we can do that with kustomize.

```
secretGenerator:
- name: app-tls
  files:
  - secret/tls.cert
  - secret/tls.key
  type: "kubernetes.io/tls"
```


https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_secretgenerator_

leifmadsen: yea I was looking at that approach, along with https://polarsquad.com/blog/loading-dynamic-configurations-in-kubernetes-kustomize which might save a few remote objects needing to be created and help with a more GitOps approach.

I'm going to experiment with both either next week while I'm in Westford, or when I return and see which approach seems to make more sense from a productization/documentation perspective.

SeanMooney: well given our intent is to store these sshkeys in a secrete use the secretGenerator would be the recommended approach or more idiomatic approach to take.

you can use env files

https://github.com/SeanMooney/nova-operator/tree/zuul-jobs-part-2/ci/nova-operator-base/topology/base

or config maps in general to make either entire fields or individual fields available in the customize context but if you intend to store files in a secret or configmap at the end of the customization then our default choice should be the ConfigmapGenerator or SecretGenerator that are part fo customize.

if we need something else then we can reach for more advanced concepts

using the env file approach vules.yamls would have to be converted to env format but it would look like this

```
configMapGenerator:
- name: environment-variables
  envs: [vaules.env]
  behavior: create

replacements:
- source:
    fieldPath: data.OPENSTACK_OPERATOR_INDEX_URL
    kind: ConfigMap
    name: environment-variables
    version: v1
  targets:
  - fieldPaths:
    - spec.image
    select:
      group: operators.coreos.com
      kind: CatalogSource
      name: openstack-operator-index
      namespace: openstack
      version: v1alpha1
```

where the env file has 
```
OPENSTACK_OPERATOR_INDEX_URL=quay.io/openstack-k8s-operators/openstack-operator-index:latest
```

anyway we can come back to this in a follow up pr

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -1,89 +1,19 @@
...
-#   CPU core isolation for your environment
-# - Change "spec.nodeTemplate.ansible.ansibleVars.edpm_neutron_sriov_agent_SRIOV_NIC_physical_device_mappings"
-#   to the desired mapping of a network name to your SRIOV interface
-#
-
-apiVersion: dataplane.openstack.org/v1beta1
-kind: OpenStackDataPlaneNodeSet
+# local-config: referenced, but not emitted by kustomize
+apiVersion: v1
+kind: ConfigMap
```
SeanMooney, line 3: i woudl genereal consier manualy defineing a configmap instead of using the configmap generator to be anti pattern by the way.

it certenly works but i don't thinkthis is the way kustomize is in intended to be used.
with that said i have not used kustomize extensively by my initial feel ing is this feels like we are sidetpping the inbuilt tools.

examples/va/nfv/sriov/edpm/values.yaml
abays: Thank you for crunching the numbers.  I've adjusted them.

abays:
@SeanMooney Thank you for your review.  We'll continue to iterate and improve upon the pattern as we move forward.

PR #78, opened 2024-01-19
Do not deploy GlanceAPI pods by default

fmount:
While we still apply the `Glance` `CR` during the first ctlplane deployment, this patch sets both `internal` and `external` instances to `replicas: 0`. By doing this we prevent a scenario where `Glance` is run without a supported backend, causing issues as long as `RWX` is not supported but we still apply a `split` layout by default.

We should still `split` the deployment between `internal` and `external` by default at this stage for at least two reasons:

1. the `layout` can't be change, so if `file` is used and a single instance is deployed, it's then not possible to patch the `default` glance instance and change its `layout` (the webhooks prevent it and the only solution is to entirely delete and recreate it);

2. in the next phase (`stage5`) we're going to patch the `ctlplane` and apply a backend (e.g., `ceph`)

For any other VA/DT that is not planning to use Ceph, it's still possible to `kustomize` at this stage the `ctlplane` and apply a backend for Glance (different than Ceph, for instance  `swift`).

Finally, as per [Glance PR#394](https://github.com/openstack-k8s-operators/glance-operator/pull/394), we are not going to return an err if Pods are not deployed and attached to a given `NAD`, so we shouldn't have issues when we land this patch.

### DIFF
--- a/lib/control-plane/openstackcontrolplane.yaml
+++ b/lib/control-plane/openstackcontrolplane.yaml
@@ -45,7 +45,7 @@ spec:
       databaseInstance: openstack
       glanceAPIs:
         default:
-          replicas: 1
+          replicas: 0
           networkAttachments:
             - storage
       storageClass: _replaced_


fultonj (commented):
I agree that we want VA1 to have 0 replicas as proposed here until Ceph is deployed.

However, because `lib/control-plane/openstackcontrolplane.yaml` is shared by all VAs  which will probably want glance by default, I propose that this be implemented by updating `va/hci/kustomization.yaml` with an additional item on the `patches` list which patches this value to `0`. 

Do you want to try that?

fultonj:
> I agree that we want VA1 to have 0 replicas as proposed here until Ceph is deployed.
> 
> However, because `lib/control-plane/openstackcontrolplane.yaml` is shared by all VAs which will probably want glance by default, I propose that this be implemented by updating `va/hci/kustomization.yaml` with an additional item on the `patches` list which patches this value to `0`.
> 
> Do you want to try that?

fultonj:
I closed this accidentally

fultonj:
I'm going to make a similar change to `va/hci/kustomization.yaml` so that swift is not deployed for VA1

fmount:
So my purpose was exactly to put `replicas:0` in the shared place `lib/control-plane/openstackcontrolplane.yaml`. By doing this for each VA/DT we can propose the right backend.
In any case with this change technically `Glance` is deployed:

```
stack@osp-storage-05 install_yamls]$ oc get glance
NAME     STATUS   MESSAGE
glance   True     Setup complete
[stack@osp-storage-05 install_yamls]$ oc get glanceapi
NAME                      NETWORKATTACHMENTS   STATUS   MESSAGE
glance-default-external                        True     Setup complete
glance-default-internal                        True     Setup complete

[stack@osp-storage-05 install_yamls]$ oc get sts | grep glance
NAME                      READY   AGE
glance-default-external   0/0     87m
glance-default-internal   0/0     87m
```
so it's a matter of configuring (as we already do with VA1) and set the appropriate backend. If we deploy glance, with no backend and this configuration, it's not going to work until a backend is configured.

fultonj:
> I agree that we want VA1 to have 0 replicas as proposed here until Ceph is deployed.
> 
> However, because `lib/control-plane/openstackcontrolplane.yaml` is shared by all VAs which will probably want glance by default, I propose that this be implemented by updating `va/hci/kustomization.yaml` with an additional item on the `patches` list which patches this value to `0`.
> 
> Do you want to try that?

I misunderstood. You're disabling Glance not just because we don't yet have Ceph specifically for VA1. The backend is generally unknown and the control plane will fail to resolve unless a backend is defined for any VA but that's outside of the realm of other VAs, e.g. SRIOV is storage backend agnostic. You're making a change to the glance operator for this too so that we follow the same model for Glance as we do for Cinder volume.

Thus, I'm +2 on this. Let's see what @abays thinks.

abays (approved):
/lgtm

I think this makes sense

PR #79, opened 2024-01-22
Create MetalLB and NMState CRs in common (before NNCP CRs)

fultonj:
This patch moves the creation of the MetalLB and NMState CRs to examples/common based on the assumption that all VAs and DTs will need them. lib/deploy is divided into lib/nmstate and lib/metallb. The common README is updated with commands to confirm MetalLB is ready and NMState is available. The HCI and NFV SRIOV VAs are updated accordingly.

Without this patch running kustomize in examples/va/hci or examples/va/nfv/sriov results in a single CR containing MetalLB, NMState and NNCP resources. When all of them are created at the same time, OpenShift will try to create the NNCP CRs while the NMState CRs are still installing and produce a resource mappings not found error. Moving MetalLB and NMState to common and updating the common README to add a step to wait for them has a side effect of resolving this problem.

The workaround section of the docs directory for this issue has been removed though the workaround docs have been changed into an FAQ so that the procedure of building smaller CRs with separate components is still documented.

Jira: OSPRH-3743


lib/metallb/kustomization.yaml
```diff
@@ -3,4 +3,3 @@ kind: Component
 
 resources:
 - metallb_deploy.yaml
```
karelyatin, line 5: Not clear if there is real need(atleast not for OSPRH-3743) to split the deployment of metallb and nmstate, could have lived together and have 1 less step.

docs/faq/cr_by_components.md
```diff
@@ -1,61 +1,50 @@
...
-no matches for kind "NodeNetworkConfigurationPolicy" in version "nmstate.io/v1"
-ensure CRDs are installed first,
-resource mapping not found for name: "ostest-master-1" namespace: "openstack"
-from "control-plane.yaml": no matches for kind "NodeNetworkConfigurationPolicy"
-in version "nmstate.io/v1"
-```
-Retrying `oc apply -f contol-plane.yaml` a few seconds later is likely to
-resolve the problem.
+The `control-plane.yaml` file contains CRs for the
+`NodeNetworkConfigurationPolicy` (NNCP), the
```
karelyatin, line 11: For avoiding OSPRH-3743 we need to split NNCP and ensure those are ready ready before moving forward with control-plane setup.

karelyatin (changes requested):

docs/faq/cr_by_components.md
fultonj: I thought the issue was NNCP being created before NMState and MetalLB were ready and not from Networking CRs racing with NNCP. At least I couldn't reproduce the problem using the current patch. I could split NNCP to a separate sub directory though in a follow up.

lib/metallb/kustomization.yaml
fultonj: In an earlier version of architecture metalb and nmstate were separated and checked separately. I don't think it hurts to have them separate. 

If we want to make it one step I'd edit `architecture/examples/common` to have one directory `nmstate_metallb` with both components (lib/metallb + lib/nmstate). I didn't like the overly generic `deploy` name. Let's see what other reviewers think. I'll focus on splitting out NNCP for now.

docs/faq/cr_by_components.md
abays: I think addressing NNCP split-out in a follow-up makes sense

fultonj:
I propose we merge this PR as is and do a follow up to split NNCP out.

abays (approved):
/lgtm with the understanding that we'll do a follow-up PR to address NNCP concerns in relation to the `OpenStackControlPlane` creation

fultonj:
@karelyatin I'm going to merge this and then do a separate PR to have NNCP be created separately. I'll start the second PR today.

PR #87, opened 2024-01-29
Remove values-ci-framework.yaml

fultonj:
The values-ci-framework.yaml file was provided as an example of the how the values file should be changed if the environmnet was deployed by the ci-framework. Now that the ci-framework is able to generate its own values.yaml file [1], we no longer need to maintain this example.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1056


abays (approved):
/lgtm

PR #88, opened 2024-01-31
Move NNCP to a separate kustomization in HCI & SRIOV VAs

fultonj:
Create nncp sub-directory for the HCI and SRIOV VAs so that the NNCPs can be created in a separate step before creating other resources which depend on them. Without this PR there are intermittent errors that "specified interfaces used to announce LB IP don't exist".

Co-Authored-By: ykarel@redhat.com
Jira: OSPRH-3743


leifmadsen:
Something about having the same contents in both VAs seems odd to me. I'm still coming up to speed on kustomize and its idiosyncrasies but having to copy the values.yaml and kustomize for nncp/ in both VAs seems like it should somehow land further down in a more common location. Can you explain why having the same NamespaceTransformer is in both VAs?

cjeanner:
From the ci-framework, I don't see any hard issue. With the [flexible loop](https://github.com/openstack-k8s-operators/ci-framework/pull/1087) implementation of the doc, we should be able to consume any directories, in the wanted order.

That said: it would be really, really good if the `edpm-pre-ceph` and `edpm-post-ceph` directories in `hci` could have the same features as the `edpm` directory nested in `nfv/sriov` with proper `values.yaml` and `kustomize.yaml`.. That's really a blocker for us for the automation.

abays (approved):
/lgtm

fultonj:
> From the ci-framework, I don't see any hard issue. With the [flexible loop](https://github.com/openstack-k8s-operators/ci-framework/pull/1087) implementation of the doc, we should be able to consume any directories, in the wanted order.
> 
> That said: it would be really, really good if the `edpm-pre-ceph` and `edpm-post-ceph` directories in `hci` could have the same features as the `edpm` directory nested in `nfv/sriov` with proper `values.yaml` and `kustomize.yaml`.. That's really a blocker for us for the automation.

https://github.com/openstack-k8s-operators/architecture/pull/91

PR #90, opened 2024-02-01
Creating `controlplane-values` config map

rebtoor:
Since storageClass parameter isn't related to networking, i'm proposing creating a new configmap named `controlplane-values` that should be used in the future for further customizations.

* linting fixes

### DIFF
--- a/examples/va/hci/kustomization.yaml
+++ b/examples/va/hci/kustomization.yaml
@@ -8,4 +8,5 @@ components:
   ## https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
 
 resources:
+  - values.yml
   - nncp/values.yaml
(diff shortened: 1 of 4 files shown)


fultonj (commented):
> "Since storageClass parameter isn't related to networking"

I was hoping that one values file for the control plane and a second for the data plane would keep things simpler. Though the values CM is called `network-volumes` maybe it should be changed to `control-plane` values.

Maybe this is OK, but I'd like to re-evaluate this proposal in after we look at another PR that @abays will be sending in today to update this repo to support edpm-{pre,post}-ceph.

fultonj:
This refactoring is not necessary for GA and there has not been activity since February. Thus, I'm closing this. Feel free to re-open and we can discuss more however.

PR #92, opened 2024-02-03
Add missing storageClass to Glance

fultonj:
Change storageClass used by Glance from "" to "host-nfs-storageclass". This makes the update to OpenStackControlPlane consistent with the original OpenStackControlPlane produced by kustomize (assuming that "host-nfs-storageclass" is used as in the default values.yaml).

Glance needs a PV from a storageClass for image staging regardless of if an external backend like Ceph RBD is used.

If we do not make this change and follow the README with the current defaults in values.yaml then the
glance-opertator complains that a forbidden change is being attempted in a StatefulSet.


abays (approved):
/lgtm

fmount (approved):
/lgtm

fmount:
Thanks John!

PR #96, opened 2024-02-09
Use nic aliases in the edpm network template

pablintino:
To allow easier automation and avoid modifying the os-net-config template of each VA to just change the interface name of the bridge change the explicit interface name to an alias.
This requires to add the os-net-config interfaces alias table to ensure proper translation between the alias and the interface name.


examples/va/hci/edpm-pre-ceph/values.yaml
```diff
@@ -27,6 +27,9 @@ data:
         edpm_iscsid_image: '{{ registry_url }}/openstack-iscsid:{{ image_tag }}'
         edpm_logrotate_crond_image: '{{ registry_url }}/openstack-cron:{{ image_tag }}'
         edpm_network_config_hide_sensitive_logs: false
+        edpm_network_config_os_net_config_mappings:
+          edpm-copmpute:
```
rebtoor, line 31: typo

examples/va/nfv/sriov/edpm/values.yaml
```diff
@@ -29,11 +29,14 @@ data:
...
         # CPU pinning settings
         edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=16 intel_iommu=on iommu=pt isolcpus=4-23,28-47"
-        edpm_tuned_profile: "cpu-partitioning-powersave" 
+        edpm_tuned_profile: "cpu-partitioning-powersave"
         edpm_tuned_isolated_cores: "4-23,28-47"
         # edpm_network_config
         # These vars are edpm_network_config role vars
         edpm_network_config_hide_sensitive_logs: false
+        edpm_network_config_os_net_config_mappings:
+          edpm-copmpute:
```
rebtoor, line 38: typo

rebtoor (changes requested):

examples/va/hci/edpm-pre-ceph/values.yaml
pablintino: Isn't [that](https://github.com/openstack-k8s-operators/edpm-ansible/blob/23473519a2b52515a4d4bdb04d264828774087c9/roles/edpm_network_config/tasks/network_config.yml#L105-L106) the name of the var?  And [this](https://github.com/openstack-k8s-operators/edpm-ansible/blob/23473519a2b52515a4d4bdb04d264828774087c9/plugins/modules/edpm_os_net_config_mappings.py#L119-L133) is how it's handled.

rebtoor (approved):

fultonj:
@pablintino would you please update this to apply the same change to the OVS-DPDK VA? 

I'm going to test it for VA1 and I'll update the review with the results

Jaganathancse (commented):
@fultonj 
Interface name to os-net-config nicX abstraction can be done using either system-uuid or system product name.
Missing to specify system uuid or system product name here
Please check link:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/9270f75ffb26dae12b83ea75adbe30f529f36d16/plugins/modules/edpm_os_net_config_mappings.py#L65
MAC address directly instead of interface name in mapping if we dont want to use either system-uuid or system product id.

fultonj:
> @fultonj Interface name to os-net-config nicX abstraction can be done using either system-uuid or system product name. Missing to specify system uuid or system product name here Please check link for example format: https://github.com/openstack-k8s-operators/edpm-ansible/blob/9270f75ffb26dae12b83ea75adbe30f529f36d16/plugins/modules/edpm_os_net_config_mappings.py#L65 MAC address directly instead of interface name in mapping if we dont want to use either system-uuid or system product id.

I see your comment @Jaganathancse but I'm not sure what the appropriate action is (if any). If there's something missing that you need perhaps you could send a PR in for it?

Jaganathancse:
> > @fultonj Interface name to os-net-config nicX abstraction can be done using either system-uuid or system product name. Missing to specify system uuid or system product name here Please check link for example format: https://github.com/openstack-k8s-operators/edpm-ansible/blob/9270f75ffb26dae12b83ea75adbe30f529f36d16/plugins/modules/edpm_os_net_config_mappings.py#L65 MAC address directly instead of interface name in mapping if we dont want to use either system-uuid or system product id.
> 
> I see your comment @Jaganathancse but I'm not sure what the appropriate action is (if any). If there's something missing that you need perhaps you could send a PR in for it?

Ack. will do it.

PR #97, opened 2024-02-09
Add automation directory

fultonj:
The automation directory contains YAML files which define the stages for each VA/DT including the path to the CR kustomization and its values as well as a shell command which may be used to validate that the stage is complete.

These YAML files may be parsed by external tools which apply the CRs in stages to automate the full deployment.

Because the structure of the architecture may change the automation directory is separate and at the top of the repository so that it does not need to change with the rest of the architecture repo.

### DIFF
--- a/automation/vars/default.yaml
+++ b/automation/vars/default.yaml
@@ -1,0 +1,56 @@
+---
+vas:
+  hci:
+    repo: https://github.com/openstack-k8s-operators/architecture.git
+    branch: main
+    stages:
+      - path: examples/va/hci/control-plane/nncp
+        validations:
+          - >-
+            oc -n openstack wait nncp
+            -l osp/nncm-config-type=standard
+            --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured
+            --timeout=60
+        values:
+          - name: network-values
+            src_file: values.yaml
+        build_output: nncp.yaml
+
+      - path: examples/va/hci/control-plane
+        validations:
+          - >-
+            oc -n openstack wait osctlplane controlplane --for condition=Ready
+            --timeout=600s
+        values:
+          - name: network-values
+            src_file: none
+        build_output: ../control-plane.yaml
+
+      - path: examples/va/hci/edpm-pre-ceph
+        validations:
+          - >-
+            oc -n openstack wait
+            osdpd edpm-deployment-pre-ceph --for condition=Ready
+            --timeout=1200s
+        values:
+          - name: edpm-values
+            src_file: values.yaml
+        build_output: dataplane-pre-ceph.yaml
+        post_stage_run:
+          - name: Deploy Ceph
+            type: playbook
+            source: "../../playbooks/ceph.yml"
+            inventory: "${HOME}/ci-framework-data/artifacts/zuul_inventory.yml"
+
+      - path: examples/va/hci
+        validations:
+          - >-
+            oc -n openstack wait
+            osdpd edpm-deployment-post-ceph --for condition=Ready
+            --timeout=1200s
+        values:
+          - name: service-values
+            src_file: service-values.yaml
+          - name: edpm-values-post-ceph
+            src_file: values.yaml
+        build_output: dataplane-post-ceph.yaml
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ Deployment topologies are only to be used for testing environments. Validated
 architectures are intended to represent production-style deployment
 environments.
 
+The [automation](automation) directory contains YAML files which
+define the stages for each VA/DT including the path to the CR
+kustomization and its values as well as a shell command which may
+be used to validate that the stage is complete.
+
 ## Requirements
 
 The templating provided here requires [kustomize](https://kustomize.io/)


leifmadsen:
So I think we could actually have a single file and not make it a set of subdirs in the automation path. How about something just like:

`automation/vars/default.yml` which would kinda match a standard Ansible vars path? Then we just need to keep extending a single file, and you can call via `-e va=hci` or whatever VA you want to run through the stages.

Looking at the POC, we'd just add:

```yaml
    - name: Include variables
      ansible.builtin.include_vars: automation/vars/default.yaml
```

Then you just need to call which VA stages you want to run.

fultonj:
> So I think we could actually have a single file and not make it a set of subdirs in the automation path. How about something just like:
> 
> `automation/vars/default.yml` which would kinda match a standard Ansible vars path? Then we just need to keep extending a single file, and you can call via `-e va=hci` or whatever VA you want to run through the stages.

OK. I have done this in the latest patch set.

leifmadsen (approved):

abays (approved):
/lgtm

leifmadsen:
Thanks John!

automation/vars/default.yaml
```diff
@@ -0,0 +1,46 @@
...
+      validation: "oc wait osdpd edpm-deployment-pre-ceph --for condition=Ready --timeout=1200s"
+      values:
+      - name: edpm-values
+        src_file: edpm_pre_ceph_values.yaml
+        dest_file: values.yaml
+
+    # Install Ceph on the OSP compute nodes in this step. OpenStack
+    # K8S CRDs do not provide a way to install Ceph via any sort of
+    # combination of CRs.
+    - path: ""
```
cjeanner, line 31: To get Ceph installation done, and to allow some more flexibility in a near future, I'd propose to add `pre_stage_run` and `post_stage_run` entries.

With this, we'd get something like:
```YAML
vas:
  hci:
    stages:
      - path: aaaa
        validation: foo bar baz
       post_stage_run:
         - name: Deploy Ceph
           type: playbook
           source: /path/to/playbooks/ceph.yml
           inventory: /path/to/zuul_inventory.yml
        values:
          - name: edpm-values
            src_file: ...
            dest_file: ....
```

And as for `pre_stage_run`, we may be able to query about Ceph secrets, unless we properly update the `ceph.yml` playbook to inject those data directly.

The structure of `pre_stage_run` and `post_stage_run` would be the same as the hooks mechanism nested in the ci-framework.
Note that it may be good to use some jinja2 parameters when it comes to hook paths (such as `source` and `inventory`) to allow the same flexibility we have in the framework regarding code location.

With that added feature, and some more tweaking to this PR while I'm integrating the automation in the framework, we should be on a safe path even for future VA (or, even DT, provided we'll provide automation for them as well).

Any thoughts?

automation/vars/default.yaml
```diff
@@ -0,0 +1,46 @@
...
+      values:
+      - name: network-values
+        src_file: network_values.yaml
+        dest_file: values.yaml
+
+    - path: examples/va/hci/control-plane
+      validation: "oc wait osctlplane controlplane --for condition=Ready --timeout=600s"
+      values:
+      - name: network-values
+        src_file: network_values.yaml
```
cjeanner, line 18: that "src_file" value is pointing to a non-existing file in the architecture tree. IIRC it was pointing to a static file in Andrew's PoC.

Would it be possible to ensure `src_file` points to the correct source file we have to modify?
Doing so, we'd be able to:
- copy that original file if we don't have done it already (ensure we're able to consume a "vanilla" original file)
- run the ci_gen_kustomize_values against it
- ensure we output the updated file to the `src_file` (override it)

automation/vars/default.yaml
```diff
@@ -0,0 +1,46 @@
...
+        dest_file: nncp/values.yaml
```
cjeanner, line 19: IIRC, `dest_file` is to override the values.yaml in Andrew's PoC. Would it be possible to point the file that should be created by `kustomize build` instead? Doing so, we'd be able to get an easy path to automation, with the `src_file` pointing to the existing values.yaml, and `dest_file` being the output needed by the `oc apply`.

automation/vars/default.yaml
```diff
@@ -0,0 +1,46 @@
...
+    branch: main
+    stages:
+    - path: examples/va/hci/control-plane/nncp
+      validation: "oc wait nncp --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured --timeout=60" 
+      values:
+      - name: network-values
+        src_file: network_values.yaml
+        dest_file: values.yaml
+
+    - path: examples/va/hci/control-plane
```
cjeanner, line 14: this one looks like a duplication, only diff being the `validation` I think?

automation/vars/default.yaml
```diff
@@ -0,0 +1,46 @@
+---
+vas:
+  hci:
+    repo: https://github.com/openstack-k8s-operators/architecture.git
+    branch: main
+    stages:
+    - path: examples/va/hci/control-plane/nncp
+      validation: "oc wait nncp --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured --timeout=60"
```
cjeanner, line 8: We may want to ensure we pass the `-n <namespace>` in the validation, to be sure we're consuming the right namespace. Also, might be good to make it a list? So that we may iterate on it and get multiple validations. Any thoughts?

cjeanner:
After re-reading the deploy files in hci VA, and tweaking a bit the proposed automation file, I ended up with this layout.
Explanations:
- repo/branches are mostly useless in this file: we'll be consuming it from the repository, so we don't want to override it IMHO. It can be done from the ci-framework directly
- validation -> validations: in case we want to inject multiple validations, let's switch to a list, even if it has only one element. Opening for more is probably a good idea for the future.
- values: now the "src_file" points to an actual `values.yaml` (and if, for some reasons, we have to point to another file, it's possible). The path is relative the the `path` of the stage. `dest_file` has 2 roles here: if different from `src_file` then it will be used as the output of `kustomize build`; if it's the same, we won't run `kustomize build`, and wait for all of the values to be generated, then call `kustomize build` in order to generate `build_output` file. This should match the last stage of HCI deploy.
- `post_stage_run` is now introduced in order to allow easy "hooks" integration. This would allow a smooth ceph deployment integration.

Any thoughts?

```YAML
---
vas:
  hci:
    stages:
      - path: examples/va/hci/
        validations:
          - >-
            oc -n openstack wait nncp
            --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured
            --timeout=60
        values:
          - name: network-values
            src_file: control-plane/nncp/values.yaml
            dest_file: nncp.yaml

      - path: examples/va/hci/control-plane
        validations:
          - >-
            oc -n openstack wait osctlplane controlplane --for condition=Ready
            --timeout=600s
        values:
          - name: network-values
            src_file: none
            dest_file: ../control-plane.yaml

      - path: examples/va/hci/
        validations:
          - >-
            oc -n openstack wait
            osdpd edpm-deployment-pre-ceph --for condition=Ready
            --timeout=1200s
        values:
          - name: edpm-values
            src_file: edpm-pre-ceph/values.yaml
            dest_file: dataplane-pre-ceph.yaml
        post_stage_run:
          - name: Deploy Ceph
            type: playbook
            source: "{{ role_path }}/../playbooks/ceph.yml"
            inventory: "{{ cifmw_basedir }}/artifacts/zuul_inventory.yml"

      - path: examples/va/hci
        validations:
          - >-
            oc -n openstack wait
            osdpd edpm-deployment-post-ceph --for condition=Ready
            --timeout=1200s
        values:
          - name: service-values
            src_file: service-values.yaml
            dest_file: service-values.yaml
          - name: edpm-values-post-ceph
            src_file: values.yaml
            dest_file: values.yaml
        build_output: dataplane-post-ceph.yaml

```

automation/vars/default.yaml
abays: We could do that and then also get rid of `dest_file`, if the idea is just to write the `src_file` back to the same location.

automation/vars/default.yaml
abays: We have separated the networking and `OpenStackControlPlane` configuration into two different steps due to errors that can occur when both are run in one `oc apply` batch.  Thus the first stage is for the networking and the second stage for the `OpenStackControlPlane`.  What might be confusing is that they both use the same `values.yaml`, but this is necessary because both need to access the networking data.

automation/vars/default.yaml
abays: Adding the namespace seems prudent.  Using a list instead also seems like a good idea.

automation/vars/default.yaml
cjeanner: `dest_file` would then be consumed in order to know "where" to output the `kustomize build` result

automation/vars/default.yaml
cjeanner: Indeed, saw it while reading the ctplane deployment doc - no problem then.

automation/vars/default.yaml
abays: Well, it wouldn't matter where you put the `kustomize build` result.  The resulting output of that is completely self-contained, so you could place wherever you wish and just `oc apply -f` it.

automation/vars/default.yaml
abays: I would say this makes sense as long as we consider this to be framework-agnostic.  The current suggestion doesn't seem to be CIFMW-specific (even though the example provided contains values that obviously work for that context, but that's fine).

automation/vars/default.yaml
abays: Yes, that's a fair point

automation/vars/default.yaml
fultonj: When Ceph is deployed by the ci-framework and the following PR has merged, then you'll have a post-ceph values.yaml file. There should not be any need to mention Ceph secrets on the `automation/vars/default.yaml`. Perhaps just the source `/home/zuul/architecture/examples/va/hci/values.yaml` which is specific to the architecture repo (`cifmw_ceph_client_values_post_ceph_path_src`)

https://github.com/openstack-k8s-operators/ci-framework/pull/1130

fultonj:
This update makes three changes:

- Change `validation` to a list of `validations`
- Replace `dest_file` under `values` list with `build_output` which is a separate item (in post ceph section we have >1 values file but each produces one build output)
- Add `post_stage_run` to account for things which need to be done outside of CRs, in this case we're using it to deploy ceph like a 3rd party (ci-framework mention could be abstracted away later but it could refer to any playbook elsewhere)

automation/vars/default.yaml
```diff
@@ -0,0 +1,56 @@
...
+    repo: https://github.com/openstack-k8s-operators/architecture.git
+    branch: main
+    stages:
+      - path: examples/va/hci/control-plane/nncp
+        validations:
+          - >-
+            oc -n openstack wait nncp
+            -l osp/nncm-config-type=standard
+            --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured
+            --timeout=60
```
cjeanner, line 13: ah, you missed the `s` for the time unit :).

abays: Yes, that is my fault for an initial bad copy-paste.  Will submit a PR.

PR #100, opened 2024-02-12
os-net-config nic alias mapping changes

Jaganathancse:
We can follow any method of below mentioned for NIC Mapping,
1. NIC Mapping based on NIC number and MAC address.
2. NIC Mapping based on NIC number and interface name using system uuid.
3. NIC Mapping based on NIC number and interface name using system product name.

### DIFF
--- a/examples/va/nfv/ovs-dpdk/edpm/values.yaml
+++ b/examples/va/nfv/ovs-dpdk/edpm/values.yaml
@@ -29,6 +29,7 @@ data:
           - hostname: clock.redhat.com
         # CPU pinning settings
         # edpm nfv ovs dpdk config
+        # CHANGEME
         edpm_kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=64 iommu=pt intel_iommu=on tsx=off isolcpus=2-11,14-23"
         edpm_tuned_profile: "cpu-partitioning-powersave"
         edpm_tuned_isolated_cores: "2-11,14-23"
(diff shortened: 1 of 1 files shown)


examples/va/nfv/ovs-dpdk/edpm/values.yaml
```diff
@@ -43,10 +44,15 @@ data:
...
         edpm_network_config_hide_sensitive_logs: false
         edpm_network_config_os_net_config_mappings:
-          edpm-compute:
-            nic2: eno2
-            nic3: eno3
-            nic4: eno4
+          # Need to provide nic mapping based on system uuid or system product name
+          # here used systetm uuid
+          edpm-compute-0: # CHANGEME
+            dmiString: "system-uuid" # CHANGEME
```
pablintino, line 50: Are `dmiString` and `id` mandatory in this case? If not, should the user remove them before applying?
@rebtoor fyi

Jaganathancse: @rebtoor yes madatory this case if we want to use interface name. otherwise need to use MAC address and NIC number mapping.

abays: How do you acquire the `id` for use with this?

Jaganathancse: Used command for uuid and product name,
# dmidecode --string system-uuid
`# dmidecode --string system-product-name`

abays: Ack, thank you @Jaganathancse

abays (approved):
/lgtm

fultonj (commented):
If this works for the VA for ovs-dpdk, then I'm fine with it. I'd rather not apply this change to the VA for HCI as it seems unnecessary.

examples/va/nfv/ovs-dpdk/edpm/values.yaml
rebtoor: @Jaganathancse care to send out a doc patch to add those commands to `os-net-config-mapping` module documentation [1] ?

[1] https://openstack-k8s-operators.github.io/edpm-ansible/collections/osp/edpm/edpm_os_net_config_mappings_module.html#ansible-collections-osp-edpm-edpm-os-net-config-mappings-module

PR #101, opened 2024-02-13
Configure OVN Controller to act as gateway

karelyatin:
To configure nicMappings(external net-attach-def) for ovnController CR, it requires dedicated nic with
external connectivity on the worker nodes.

With network isolation we already setup an additional nic on the worker nodes and as that is already used by MetalLB and net-attach-definitions it cannot be
shared by OVN Controllers as it uses host-device
plugin and requires a dedicated NIC.

In order to use same interface for ovn controller, a linux bridge is created with with ctlplane interface as one of it's port. Utilizing this bridge a net-attach-def of bridge type is created for usage by the OVN Controllers.

If there is additional external NIC dedicated for OSP traffic on worker nodes then that can also be used.

Issue: OSPRH-3947

### DIFF
--- a/examples/va/hci/control-plane/nncp/values.yaml
+++ b/examples/va/hci/control-plane/nncp/values.yaml
@@ -52,7 +52,7 @@ data:
         "cniVersion": "0.3.1",
         "name": "ctlplane",
         "type": "macvlan",
-        "master": "enp7s0",
+        "master": "ospbr",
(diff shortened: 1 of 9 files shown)


abays (commented):
Seems fine to me as long as https://github.com/openstack-k8s-operators/ci-framework/pull/1154 merges with it

PR #102, opened 2024-02-13
Remove references to host-nfs-storageclass

abays:
Replace references to `host-nfs-storageclass` with `local-storage` instead.  The former is an old artifact for an OCP-cluster-deployment tool that was used in the nascent VA effort.  We now use `local-storage` with all our environments, so we should just have that as a default instead.  I've noticed situations multiple times now where the old storage class trips people up when somehow it stays in place during VA deployments.


pablintino (approved):

fmount (approved):
/lgtm

PR #103, opened 2024-02-15
Add reboot-os service to hci and nfv va examples

ciecierski:
Add reboot-os service to nfv and hci validated architectures examples.
This service should be enabled by default in each architecture,
because there might be changes to host during deployment that
requires reboot to take effect.


fultonj (approved):
I see this is already in the SRIOV version.

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L145

abays (approved):
/lgtm

PR #104, opened 2024-02-15
Add missing 's' to timeout in automation file

abays:
When I initially copy-pasted the validation content in the PoC for this structure, I missed an `s` qualifier for a `timeout` parameter


cjeanner (approved):
Thanks!

leifmadsen (approved):

PR #107, opened 2024-02-19
Global MTU size default to 9000

Jaganathancse:


examples/va/nfv/ovs-dpdk/service-values.yaml
```diff
@@ -16,6 +16,8 @@ metadata:
 data:
   neutron:
     customServiceConfig: |
+      [DEFAULT]
+        global_physnet_mtu = 9000
```
abays, line 20: Indentation needs to be fixed, I think

Jaganathancse: done

PR #108, opened 2024-02-19
OVN Controller bridge mapping to enable floating ip

Jaganathancse:
This change is to provide external connectivity for the guest instances.


abays (approved):
/lgtm

PR #109, opened 2024-02-27
OVS DPDK & SRIOV dataplane service order changes

Jaganathancse:
This change is to change the openstack dataplane
services order.


abays (approved):
/lgtm

PR #110, opened 2024-02-27
Remove "src_file" when none is provided

cjeanner:
It's better/easier to ensure the "src_file" property is defined rather
than checking for "none" as a string, like ansible is doing against that
file.

### DIFF
--- a/automation/vars/default.yaml
+++ b/automation/vars/default.yaml
@@ -21,7 +21,6 @@ vas:
             --timeout=600s
         values:
           - name: network-values
-            src_file: none
         build_output: ../control-plane.yaml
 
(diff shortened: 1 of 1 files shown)


cjeanner:
@fultonj @abays this is something that should make the whole structure easier to parse/check. We're already doing the "right" check in our open [PR](https://github.com/openstack-k8s-operators/ci-framework/pull/1168/files#diff-888ede7c7622d73459cf89a99ce06525e383be568e73d2fed5453eff970810a8R20) - we'll clean the other "checks" once this src_file is removed.

fultonj (approved):

PR #82, opened 2024-01-26
CRs for unified delta deployed topology

psathyan:
This PR contains the necessary custom resources for deploying OpenStack Services on OpenShift. The CRs are for the proposed uni*delta deployed topology.

PR #111, opened 2024-03-04
[WIP] [DNM] Add DT to support external ceph deployment

rachael-george:
OSPRH-2737 - Support external Ceph deployment


tosky:
There is an DT which tries to deploy on external ceph. As we want to keep the amount of DT minimum, please try to join that effort.
https://github.com/openstack-k8s-operators/architecture/pull/82

PR #112, opened 2024-03-04
Switch glance replicas to 3 in HCI VA

fultonj:
Glance distributed image import using the glance operator's stateful set has been tested and is known to work with three replicas. Increase replica count in HCI VA to so that when Glance is tested it will be tested with three replicas.

### DIFF
--- a/examples/va/hci/service-values.yaml
+++ b/examples/va/hci/service-values.yaml
@@ -38,7 +38,7 @@ data:
       rbd_store_user = openstack
     glanceAPIs:
       default:
-        replicas: 1
+        replicas: 3
   manila:
     enabled: true
     manilaShares:


abays (approved):
/lgtm

fmount (approved):

tosky:
I thought we were using RGW in VA1. It would make sense given the "all ceph" deployment. If it's not the case, can we please revisit the decision?

fultonj:
> I thought we were using RGW in VA1. It would make sense given the "all ceph" deployment. If it's not the case, can we please revisit the decision?

This patch is about 3 replicas using glance RBD. 

We also deploy RGW for object store (and can configure it via ci framework in a follow up) but what does that have to do with Glance replicas?

VA1 should use RBD for cinder, glance, nova for COW boots. 

I.e. if you're asking for Glance to use RGW instead of RBD, then I don't think it should in VA1. But regardless of backed, distributed image import is about referencing a non-shared staging area. The discussion of glance backed does not need to block distributed image import or replicas.

tosky:
> > I thought we were using RGW in VA1. It would make sense given the "all ceph" deployment. If it's not the case, can we please revisit the decision?
> 
> This patch is about 3 replicas.
> 
> VA1 should use RBD for cinder, glance, nova for COW boots.

I apologize. I mixed up two reviews and I read "swift" instead of "glance" here. Please ignore my comment.

PR #113, opened 2024-03-05
Fix title of dataplane docs for the VAs

abays:
Copy-paste mistake


examples/va/hci/dataplane-post-ceph.md
```diff
@@ -1,4 +1,4 @@
-# Configuring networking and deploy the OpenStack control plane
+# Configure and deploy the post-Ceph dataplane
```
leifmadsen, line 1: Generally titles should have gerunds, so a better title would be:

> Configuring and deploying the post-Ceph dataplane environment

leifmadsen (changes requested):

leifmadsen (approved):

PR #115, opened 2024-03-06
Added the `openstack-operator-image` variable

rebtoor:
Added the ability to customize the "release" of the openstack-operator (index) image to deploy a different set of containers instead of the ones pinned in "latest"

Defaults to
`quay.io/openstack-k8s-operators/openstack-operator-index:latest`


fultonj (commented):
+2 
This looks OK to me. Though, I'd prefer @abays review and approve.

pablintino:
> This fine with me for now, as we already overload `network-values` with another item (`storageClass`). But in the future we need to separate these non-network values.
> 
> /lgtm

Agree, this kind of information deserves another place.

PR #116, opened 2024-03-06
Move `openstack-operator-version` variable to `olm-values` CM

rebtoor:
This is a followup to #115 since that approach failed while deploying OLM resources.

Also, `openstack-operator-version` variable has been moved to a new CM called `olm-values`


fultonj (commented):
+2

abays (approved):
/lgtm

PR #117, opened 2024-03-11
Removes openstack-operator-image from network-values configmap

frenzyfriday:
As per https://github.com/openstack-k8s-operators/architecture/pull/116 source of replacement for `data.openstack-operator-image` is moved to olm-values  from network-values


abays (approved):
/lgtm

PR #120, opened 2024-03-12
Missing glance and swift controlplane service values

Jaganathancse:
This change is to add the missing glance and swift controlplane service values for OVS DPDK deployment.


abays (approved):
/lgtm

PR #121, opened 2024-03-12
Missing NIC relative name

Jaganathancse:
Need to use relative name instead of interface name in the os-net-config template.


abays (approved):
/lgtm

PR #127, opened 2024-03-18
EDPM Compute NIC Mapping changes

Jaganathancse:
This is to update os-net-config NIC mapping with
NIC numbering and MAC address.


abays (approved):
/lgtm

PR #128, opened 2024-03-18
HCI EDPM compute NIC Mapping changes

Jaganathancse:
This is to update os-net-config NIC Mapping issue
looks like default NIC mapping os-net-config is used.

### DIFF
--- a/examples/va/hci/edpm-pre-ceph/values.yaml
+++ b/examples/va/hci/edpm-pre-ceph/values.yaml
@@ -28,8 +28,12 @@ data:
         edpm_logrotate_crond_image: '{{ registry_url }}/openstack-cron:{{ image_tag }}'
         edpm_network_config_hide_sensitive_logs: false
         edpm_network_config_os_net_config_mappings:
-          edpm-compute:
-            nic2: enp7s0
+          edpm-compute-0:
+            nic2: 6a:fe:54:3f:8a:02 # CHANGEME
+          edpm-compute-1:
+            nic2: 6b:fe:54:3f:8a:02 # CHANGEME
+          edpm-compute-2:
+            nic2: 6c:fe:54:3f:8a:02 # CHANGEME
         edpm_network_config_template: |
           ---
           {% set mtu_list = [ctlplane_mtu] %}


fultonj (changes requested):
I see two changes:

1. edpm-compute to edpm-compute-i for {0,1,2}
2. enp7s0 to $MAC_ADDRESS

If we need to switch 1, could we do only that in this change and not do the mac address change?

Jaganathancse:
> 1. edpm-compute-i

@fultonj 

edpm-compute-i:
    nic2: enp7s0 

This is only needed?
Finding mac address for the interface on all the compute is done via automation script?

Jaganathancse:
> I see two changes:
> 
> 1. edpm-compute to edpm-compute-i for {0,1,2}
> 2. enp7s0 to $MAC_ADDRESS
> 
> If we need to switch 1, could we do only that in this change and not do the mac address change?

If we need to use interface name in mapping, need to specify machine UUID or system-product-name

Jaganathancse:
> I see two changes:
> 
> 1. edpm-compute to edpm-compute-i for {0,1,2}
> 2. enp7s0 to $MAC_ADDRESS
> 
> If we need to switch 1, could we do only that in this change and not do the mac address change?

using interface name with nic numbering mapping is not allowed.
need to use both. otherwise dmiString logic to be used.

fultonj:
> > I see two changes:
> > 
> > 1. edpm-compute to edpm-compute-i for {0,1,2}
> > 2. enp7s0 to $MAC_ADDRESS
> > 
> > If we need to switch 1, could we do only that in this change and not do the mac address change?
> 
> using interface name with nic numbering mapping is not allowed. need to use both. otherwise dmiString logic to be used.

Why do we need this change at all? In the commit message it says "update os-net-config NIC Mapping issue". Can I find out more details about this? At least for me the current defaults are working fine so I don't see why we need to make this change but if something is broken, I'd like to understand what.

examples/va/hci/edpm-pre-ceph/values.yaml
```diff
@@ -28,8 +28,12 @@ data:
...
         edpm_network_config_hide_sensitive_logs: false
         edpm_network_config_os_net_config_mappings:
-          edpm-compute:
-            nic2: enp7s0
+          edpm-compute-0:
+            nic2: 6a:fe:54:3f:8a:02 # CHANGEME
+          edpm-compute-1:
+            nic2: 6b:fe:54:3f:8a:02 # CHANGEME
+          edpm-compute-2:
+            nic2: 6c:fe:54:3f:8a:02 # CHANGEME
```
fultonj, line 36: I talked to Andrew and I understand better what is happening here.

* Lines 31 and 32 are not doing anything and `os-net-config` was ignoring them and just using the defaults
* The defaults worked in a lot of testing but fail in some cases (which this PR is trying to fix with an example)
* The defaults can be corrected by using the newly proposed syntax, provided the mac addressees are correct
* The alternative to using mac addresses is to provide DMI but what would be more difficult for ci-framework to provide in the values.yaml file it will replace when it overrides this example

So I'm fine with this example being changed since I now understand why.

I expect ci-framework has the desired mac address stored as a fact so maybe it could generate a values.yaml like this example.

fultonj (approved):
After talking to Andrew I understand better why this change is proposed.

PR #129, opened 2024-03-19
Add cephfs auth_id for manila in VA1

fmount:
This patch fixes the `Manila` `native_cephfs` config for `VA1`.
A missing `cephfs_auth_id` let the share fail because it tries to authenticate with a different user.
The patch also adds the `manilaAPI` `customServiceConfig` to make sure we're able to handle that protocol at `API` level.

### DIFF
--- a/examples/va/hci/service-values.yaml
+++ b/examples/va/hci/service-values.yaml
@@ -41,6 +41,10 @@ data:
         replicas: 3
   manila:
     enabled: true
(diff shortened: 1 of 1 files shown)


raukadah:
lgtm, tested in the HCI deployment. Thank you @fmount for fixing it. :-)

abays (approved):
/lgtm

PR #132, opened 2024-03-19
EDPM Compute with provision changes

Jaganathancse:
This change is to support EDPM Compute with provision changes.

### DIFF
--- a/examples/va/nfv/ovs-dpdk-sriov/edpm/values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/edpm/values.yaml
@@ -6,6 +6,13 @@ metadata:
   annotations:
     config.kubernetes.io/local-config: "true"
 data:
+  preProvisioned: false
+  baremetalSetTemplate:
+    ctlplaneInterface: eno2 #CHANGEME
+    cloudUserName: cloud-admin
+    provisioningInterface: enp1s0 #CHANGEME
+    bmhLabelSelector:
+      app: openstack #CHANGEME
   ssh_keys:
     # Authorized keys that will have access to the dataplane computes via SSH
     authorized: CHANGEME
--- a/examples/va/nfv/ovs-dpdk/edpm/values.yaml
+++ b/examples/va/nfv/ovs-dpdk/edpm/values.yaml
@@ -6,6 +6,13 @@ metadata:
   annotations:
     config.kubernetes.io/local-config: "true"
 data:
+  preProvisioned: false
+  baremetalSetTemplate:
+    ctlplaneInterface: eno2 #CHANGEME
+    cloudUserName: cloud-admin
+    provisioningInterface: enp1s0 #CHANGEME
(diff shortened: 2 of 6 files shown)


fultonj (commented):
I've done all my testing with preprovisioned EDPM nodes and that's also the ci-framework default.

Could we default preProvisioned to true? and have the kustomization switch it to false if it's set to false in the values.yaml?

 Similarly, could `baremetalSetTemplate` be added by kustomize when applicable?

@abays please let me know if I'm overlooking something.

Jaganathancse:
> I've done all my testing with preprovisioned EDPM nodes and that's also the ci-framework default.
> 
> Could we default preProvisioned to true? and have the kustomization switch it to false if it's set to false in the values.yaml?
done
> 
> Similarly, could `baremetalSetTemplate` be added by kustomize when applicable?
> 
done
> @abays please let me know if I'm overlooking something.

Thanks @fultonj for comments. Please review the updated changes.

fultonj (approved):
Thanks @Jaganathancse 

@abays are you OK with this?

abays:
I think it would be better to continue to use the `values*.yaml` pattern than introduce a new pattern where the user has to modify `kustomization.yaml`.  Could we move the fields marked `# CHANGEME` into the `values.yaml` and change the `patches` into `replacements` to retain the established pattern?

abays:
> I think it would be better to continue to use the `values*.yaml` pattern than introduce a new pattern where the user has to modify `kustomization.yaml`. Could we move the fields marked `# CHANGEME` into the `values.yaml` and change the `patches` into `replacements` to retain the established pattern?

i.e. add the `preProvisioned` field and value along with the `baremetalSetTemplate` field and value to the respective `values.yaml` and then do something like [1], except make those changes in `examples/va/nfv/.../edpm/kustomization.yaml` instead of `va/nfv/.../edpm/kustomization.yaml`.

[1] https://github.com/openstack-k8s-operators/architecture/blob/6af47de32df784a87885d0a0f38bbe91c2abbab6/va/nfv/sriov/edpm/kustomization.yaml#L25-L51

Jaganathancse:
> > I think it would be better to continue to use the `values*.yaml` pattern than introduce a new pattern where the user has to modify `kustomization.yaml`. Could we move the fields marked `# CHANGEME` into the `values.yaml` and change the `patches` into `replacements` to retain the established pattern?
> 
> i.e. add the `preProvisioned` field and value along with the `baremetalSetTemplate` field and value to the respective `values.yaml` and then do something like [1], except make those changes in `examples/va/nfv/.../edpm/kustomization.yaml` instead of `va/nfv/.../edpm/kustomization.yaml`.
> 
> [1]
> 
> https://github.com/openstack-k8s-operators/architecture/blob/6af47de32df784a87885d0a0f38bbe91c2abbab6/va/nfv/sriov/edpm/kustomization.yaml#L25-L51

done

PR #136, opened 2024-03-21
Increase edpm-deployment-post-ceph validation timeout

eurijon:
**Note: this change is being tested right now**

Similarly to https://github.com/openstack-k8s-operators/architecture/pull/135, this patch increases the timeout for the edpm-deployment-post-ceph OpenStackDataPlaneDeployment to get Ready in VA for shiftonstack case, which includes the next overrides in ci-framework:

 - cifmw_libvirt_manager_compute_disksize: 200
 - cifmw_libvirt_manager_compute_memory: 50
 - cifmw_libvirt_manager_compute_cpus: 8
 - cifmw_block_device_size: 100G

This patch increases it from 20min (1200s) to 35min (2100s) as it's taking ~29min to get Ready in my tests.

PR #137, opened 2024-03-22
Bump control plane and dataplane deployment timeout to 30m

raukadah:
Recently we have added VA HCI ci in downstream using ci-framework devscript reproducer. Sometimes control plane or dataplane deployment does not finishes with existing timeout.

In order to finish the deployment, we are bumping the timeout to 30 m for controlplane and dataplane deployment.

In upstream Zuul CI, the controlplane deploy timeout[1] is 30 mins and EDPM deploy timeout[2] is 40 mins.

Links:
[1]. https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/edpm_prepare/defaults/main.yml#L27 

[2]. https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/edpm_deploy/defaults/main.yml#L25

### DIFF
--- a/automation/vars/default.yaml
+++ b/automation/vars/default.yaml
@@ -18,7 +18,7 @@ vas:
         validations:
           - >-
             oc -n openstack wait osctlplane controlplane --for condition=Ready
-            --timeout=600s
+            --timeout=30m
         values:
           - name: network-values
             src_file: nncp/values.yaml
@@ -45,7 +45,7 @@ vas:
           - >-
             oc -n openstack wait
             osdpd edpm-deployment-post-ceph --for condition=Ready
-            --timeout=1200s
+            --timeout=40m
         values:
           - name: service-values
             src_file: service-values.yaml


cjeanner:
this ones conflicts with #136 - might be good to sync with @eurijon ?

cjeanner:
+2
I don't see any issue with that.
@fultonj anything against this PR? it's needed for our internal CI :)

automation/vars/default.yaml
```diff
@@ -45,7 +45,7 @@ vas:
           - >-
             oc -n openstack wait
             osdpd edpm-deployment-post-ceph --for condition=Ready
-            --timeout=1200s
+            --timeout=40m
```
fultonj, line 48: In the following we bumped this to 35 minutes. I assume @eurijon does not mind if we add 5 more minutes.

  https://github.com/openstack-k8s-operators/architecture/pull/136/files

automation/vars/default.yaml
```diff
@@ -18,7 +18,7 @@ vas:
         validations:
           - >-
             oc -n openstack wait osctlplane controlplane --for condition=Ready
-            --timeout=600s
+            --timeout=30m
```
fultonj, line 21: OK, we're going from 10 minutes to 30 here. That's fine too.

fultonj (approved):

PR #138, opened 2024-03-25
default route config changes

Jaganathancse:
This change is to update default destination route config changes for NFV features.


abays (approved):
We'll try this for now but will follow-up if it causes other problems

/lgtm

PR #141, opened 2024-03-26
Openstack dataplane service nova custom name changes

Jaganathancse:
This change is to fix nova custom dataplane service naming issue for ovs dpdk sriov.


abays (approved):
/lgtm

PR #144, opened 2024-03-26
Adding baremetalset root password changes

Jaganathancse:
This change is to add baremetalset root password secret for compute nodes to debug via virtual console.

### DIFF
--- a/examples/va/nfv/ovs-dpdk-sriov/edpm/values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/edpm/values.yaml
@@ -6,6 +6,7 @@ metadata:
   annotations:
     config.kubernetes.io/local-config: "true"
 data:
+  root_password: cmVkaGF0Cg==
   preProvisioned: false
   baremetalSetTemplate:
     ctlplaneInterface: eno2 #CHANGEME
@@ -13,6 +14,9 @@ data:
     provisioningInterface: enp1s0 #CHANGEME
     bmhLabelSelector:
       app: openstack #CHANGEME
+    passwordSecret:
+      name: baremetalset-password-secret
+      namespace: openstack
   ssh_keys:
     # Authorized keys that will have access to the dataplane computes via SSH
     authorized: CHANGEME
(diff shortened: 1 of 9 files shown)


abays:
The HCI VA example needs to be updated as well or you get this error:

```
$ kustomize build edpm-pre-ceph > dataplane-pre-ceph.yaml
Error: accumulating components: accumulateDirectory: "recursed accumulation of path '/home/ocp/podified/Jaganathancse-architecture/va/hci/edpm-pre-ceph': accumulating components: accumulateDirectory: \"recursed accumulation of path '/home/ocp/podified/Jaganathancse-architecture/lib/dataplane': fieldPath `data.root_password` is missing for replacement source ConfigMap.[noVer].[noGrp]/edpm-values.[noNs]\""
```

abays:
Actually, sorry, I think it would be better to actually have the kustomization in the `va/nfv/*/` directories instead of the `lib` directory.  The reason for this is that the NFV VAs are using baremetal provisioning, whereas the HCI VA is using preprovisioned.  The baremetal secret only applies to baremetal provisioning, so generating it for the HCI VA doesn't make sense.  So maybe we should undo what I suggested earlier and move the logic out of `lib/dataplane/` and into each `va/nfv/*/` directory.

@fultonj thoughts?

fultonj (approved):
So we're going back to patch set 2?

abays:
> So we're going back to patch set 2?

We would be moving the logic out of `lib/dataplane` and into the middle-tier of `va/nfv/*/` so that the baremetal password is properly applied to only the VAs that currently use baremetal provisioning.

fultonj:
> Actually, sorry, I think it would be better to actually have the kustomization in the `va/nfv/*/` directories instead of the `lib` directory. The reason for this is that the NFV VAs are using baremetal provisioning, whereas the HCI VA is using preprovisioned. 

Yes, HCI is using pre-provisioned. In theory someone could use baremetal provisioned for HCI too. However, because we are being prescriptive here, I agree with the above.

> The baremetal secret only applies to baremetal provisioning, so generating it for the HCI VA doesn't make sense. 

agreed.

> So maybe we should undo what I suggested earlier and move the logic out of `lib/dataplane/` and into each `va/nfv/*/` directory.
> 
> @fultonj thoughts?

yes.

abays (approved):
/lgtm

PR #148, opened 2024-03-27
Network config template changes

Jaganathancse:
This change is to fix the ovs dpdk and ovs dpdk sriov network config template.


abays (approved):
/lgtm

PR #149, opened 2024-03-28
BareMetalHost CR changes

Jaganathancse:
This change is to add BareMetalHost CR support for NFV OVS DPDK EDPM deployment.


abays:
Did we decide to put the BMHs in the VAs themselves?  For CIFMW, for instance, they're indicating that they will just handle it themselves as far as I know at the moment.  I'm just not sure what is proper here.

Jaganathancse:
@abays 
This PR is not needed, bmh CR automation will be part of CIFW.

PR #153, opened 2024-04-02
telemetry, glance swift and ovn controller controlplane configs

Jaganathancse:
This change is to fix glance, swift and ovn controller controlplane custom configurations.


abays (approved):
/lgtm

PR #156, opened 2024-04-03
Add new workflow: yamllint

cjeanner:
This should ensure some consistency within the various YAML files.


abays (approved):
Tested the various `kustomize build` command across the VAs and it looks good

/lgtm

PR #157, opened 2024-04-03
Staticize dhcp config for eth0

lmiccini:
We used to rely on libvirt assigning (and keeping track) of dhcp leases to configure eth0 ip addresses.
Let's make sure we persist this into os-net-config so that we do not accidentally lose this config between reboots.

### DIFF
--- a/examples/va/hci/edpm-pre-ceph/values.yaml
+++ b/examples/va/hci/edpm-pre-ceph/values.yaml
@@ -45,6 +45,10 @@ data:
           {%- endfor %}
           {% set min_viable_mtu = mtu_list | max %}
           network_config:
+          - type: interface
+            name: nic1
+            use_dhcp: true
+            mtu: {{ min_viable_mtu }}
           - type: ovs_bridge
             name: {{ neutron_physical_bridge_name }}
             mtu: {{ min_viable_mtu }}


examples/va/hci/edpm-pre-ceph/values.yaml
```diff
@@ -45,6 +45,10 @@ data:
           {%- endfor %}
           {% set min_viable_mtu = mtu_list | max %}
           network_config:
+          - type: interface
```
cjeanner, line 48: we'll need to amend the ci-framework template in order to make the deploy aware of `nic1`:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/hci/edpm-values/values.yaml.j2#L28-L31

rebtoor (commented):
I'd add a fake mac address like 
`nic1: 6b:fe:54:3f:8a:00 # CHANGEME` for each edpm-compute in `edpm_network_config_os_net_config_mappings` section (lines 34-39)

examples/va/hci/edpm-pre-ceph/values.yaml
cjeanner: https://github.com/openstack-k8s-operators/ci-framework/pull/1390 done here

rebtoor (approved):

abays (approved):
/lgtm

PR #158, opened 2024-04-03
Add a values.yaml for OpenStackDataPlaneDeployment

fao89:


cjeanner (commented):
GIving a try right now with a scale-out scenario:
- first deploy will happen with "vanilla" name
- second deploy will add a compute and update the deployment name via a user provided "kustomization" in the ci-framework.

cjeanner:
++ It works as expected.

Note that the CI Framework needs to inject a custom wait_conditions to catch the new name, buuut... it's a thing, and that change makes automated scaling-out possible!

PR #160, opened 2024-04-03
nfv kustomization: Replace only nic Mappings

mnietoji:
When using, ovn-controller, we are overwriting other fields. The networkAttachment field is lost, we could add this field to our values files, or we could just update nicMappings. Not sure which is the best solution.

Without the pach:
  ovn:
    template:
      ovnController:
        nicMappings:
          access: enp6s0

With the patch:
  ovn:
    template:
      ovnController:
        networkAttachment: tenant
        nicMappings:
          access: enp6s0


Jaganathancse (approved):

Jaganathancse:
@mnietoji Please resolve conflicts to merge it.

Jaganathancse (approved):

PR #161, opened 2024-04-04
Add script to verify kustomizations

sdatko:
This commit introduces simple Bash script that can be used to verify whether the prepared kustomization files can be built.

### DIFF
--- a/test-kustomizations.sh
+++ b/test-kustomizations.sh
@@ -1,4 +1,12 @@
 #!/usr/bin/env bash
+#
+# This script verifies whether the prepared kustomization files can be built.
+# By default it looks for all kustomizations under current working directory,
+# but it is possible to specify desired directories as the script's arguments.
+#
+# Usage:
+#   ./test-kustomizations.sh [dir1 dir2 ...]
+#
 set -o errexit
 set -o nounset
 set -o pipefail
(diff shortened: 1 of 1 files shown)


leifmadsen:
I went looking to see if there was maybe a kustomize linter already that we could use, and ended up running into this article: https://developers.redhat.com/articles/2023/10/10/how-validate-gitops-manifests

I saw that it has a script created that basically does what you're proposing here. I wonder if it'd be better to align to the existing script maybe?

* https://developers.redhat.com/articles/2023/10/10/how-validate-gitops-manifests?source=sso#kustomize_validation
* https://github.com/redhat-cop/gitops-catalog/blob/main/scripts/validate_manifests.sh

sdatko:
I have no strong opinion here – I just needed something for my work, so I made it, then I felt I may share it, as there was nothing yet (the YAML linter appeared as GitHub actions just yesterday).

As for the mentioned script, there are some parts I do not like and consider an anti-pattern in Bash; it even looks a bit like a draft, though it may also provide support for validating schemas (after uncommenting some parts). So I would not use it directly without modifications anyway, but that is a personal opinion. However, it gives me ideas on what can I improve in my small script, so thank you for linking it – I am barely starting my journey towards Kubernetes/OpenShift (I was just told today to use `kustomize build` for basic verification of what I wrote).

cjeanner:
Maybe we can integrate this in the existing workflow, for instance the yaml-lint one (since it's triggered on any YAML file change) ? Even if we switch CI "provider", or think another "validation tool" is better, we'd get a base with your script, allowing to be more confident when reviewing and eventually merging patches?

If so, what about:
- move the script in the .ci/ directory
- add a note in the README
- call the script from within the "linter.yml" workflow (https://github.com/openstack-k8s-operators/architecture/blob/main/.github/workflows/linter.yaml)

Would that be something good?

abays (approved):
/lgtm

PR #162, opened 2024-04-04
Add new worklow: validate automation schema

cjeanner:
Using yamale, we're now ensuring the automation file schema is correct:
- check for required data
- check for data type
- check overall structure of the file


abays (approved):
/lgtm

PR #163, opened 2024-04-04
Ensure wait_conditions starts with specific pattern

cjeanner:
This ensures we're really facing a wait_condition, by checking we're
starting the command by either `oc ` or `kubectl ` (yes, with a space
after the command).


abays (approved):
/lgtm

PR #164, opened 2024-04-04
Add new CI validation: ensure files exist

cjeanner:
Since we now ensure the automation schema is valid, we also want to
ensure linked directories and files exist in the various automation
scenarios.

This new script will load any YAML file nested in the automation/vars
directory, and ensures all the `path` and `src_file` are present.

### DIFF
--- a/.ci/validate-schema-paths.py
+++ b/.ci/validate-schema-paths.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+
+import glob
+import pathlib
+import yaml
+
+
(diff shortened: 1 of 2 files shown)


cjeanner:
@abays any thoughts on that one?

I have just a small "maybe..." related to the amount of github actions we're allowed to run, either per repository or in the organzation: maybe I should get everything in one action, stacking the steps? It would make the output slightly "uglier", but we'd be able to spare on github limits... ?

PR #167, opened 2024-04-05
Add missing tls params to nova custom service

karelyatin:
It's required with tls enabled, [1] already enabled tls by default so need to include the missing certs.
The issue should be visible once dataplane-operator is updated[2] or tlsEnabled flag is explicitly set in the nodeset.

[1] openstack-k8s-operators/dataplane-operator#754
[2] openstack-k8s-operators/openstack-operator#732


cjeanner (approved):

abays (approved):
/lgtm

PR #171, opened 2024-04-10
Don't use networks both in camelcase and smallcase

rabi:
Don't use network names both in camelcase and smallcase in the same spec. Also you don't need to set them in nodeset template if provided in the node section.

Also changes networks to make them all smallcase.

### DIFF
--- a/examples/va/nfv/ovs-dpdk/edpm/values.yaml
+++ b/examples/va/nfv/ovs-dpdk/edpm/values.yaml
@@ -87,7 +87,7 @@ data:
               mtu: {{ min_viable_mtu }}
               # force the MAC address of the bridge to this interface
               primary: true
-          {% for network in nodeset_networks if network not in ['External', 'Tenant'] %}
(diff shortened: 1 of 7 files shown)


examples/va/nfv/ovs-dpdk/edpm/values.yaml
```diff
@@ -87,7 +87,7 @@ data:
               mtu: {{ min_viable_mtu }}
               # force the MAC address of the bridge to this interface
               primary: true
-          {% for network in nodeset_networks if network not in ['External', 'Tenant'] %}
+          {% for network in nodeset_networks if network not in ['External', 'tenant'] %}
```
abays, line 90: Just want to check: is it okay to still have a capitalized `External` here?

rabi: It would work as it's External in NetConfig https://github.com/openstack-k8s-operators/architecture/blob/main/lib/networking/netconfig.yaml#L20, but will change both, thanks!

abays (approved):
/lgtm

PR #175, opened 2024-04-10
Add a new test in the linter workflow

cjeanner:
This new test leverages a script introduced by #161, allowing to ensure
at least `kustomize build` works.

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -8,6 +8,7 @@ on:  # yamllint disable-line rule:truthy
     paths:
       - "**.yaml"
       - "**.yml"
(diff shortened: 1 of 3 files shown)


fultonj:
Thank you for adding a basic kustomize workflow so that we don't have patches which break the ability to run kustomize on our examples directory.

We are removing the `one` directory merely because the empty file made the job not pass and that file was just meant as an example.

```
examples/dt/one/kustomization.yaml: ERROR
Error: kustomization.yaml is empty

:: Summary
1 errors!

Error: Process completed with exit code 1.
```

fultonj (approved):

PR #177, opened 2024-04-11
remove route in nncp

mnietoji:
It is causing issues, it is not possible to connect from any pod, for example tempest pod or openstackclient pod to compute nodes using control plane network (192.168.122.0/24)

Not sure what issue is trying to solve this route, I think it is not needed

### DIFF
--- a/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
@@ -182,12 +182,6 @@ data:
         values:
           - 192.168.122.1
 
-  routes:
-    config:
-      - destination: 192.168.122.0/24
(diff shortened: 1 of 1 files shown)


fultonj:
-1 

This is failing the following test
  https://github.com/openstack-k8s-operators/architecture/actions/runs/8647363849/job/23708643221?pr=177

fultonj:
Please run `./test-kustomizations.sh examples/` in your branch and then update until it passes.

leifmadsen:
@mnietoji @karelyatin here is an example PR of a more appropriate way to address this, since removing the configmap values won't work, as they are required by the base components:  https://github.com/openstack-k8s-operators/architecture/pull/203

Do you still feel this PR is necessary? If not, feel free to close it out.

karelyatin:
Closing as per last discussion above

PR #179, opened 2024-04-15
Revert Staticize dhcp config for eth0

rabi:
Using dhcp for  nic1 breaks the dns configuration by updating resolv.conf used by edpm. Reverting to fix current issues and if there is a requirement to use dhcp for nic1 it could be reworked.


abays (approved):
Okay, let's merge this for now and follow-up with some other approach to achieve what was originally needed

/lgtm

PR #180, opened 2024-04-15
NUMA aware vSwitch vlan and vxlan nova config changes

Jaganathancse:
This changes is to move vlan and vxlan configs to nova config.


fultonj (commented):
+2

abays (approved):
/lgtm

PR #185, opened 2024-04-16
Remove nncp next-hope-interface from ConfigMap

pablintino:
All the network-values ConfigMaps has the next-hope-interface set in the route placed there. That value is not consumed anywhere as the generated NNCPs uses the bridge interface name as next-hop and not the one set there.
The presence of this field may confuse a user.


abays:
I think this can be closed because it has been superseded by https://github.com/openstack-k8s-operators/architecture/pull/208

leifmadsen:
> I think this can be closed because it has been superseded by #208

Agreed. Closing!

PR #188, opened 2024-04-19
Add missing libvirt-secret

psathyan:
With the commit in [dataplane-operator#831](https://github.com/openstack-k8s-operators/dataplane-operator/pull/831), libvirt-secret is now required to be available before nova service can be deployed on the compute nodes.

```
message: Deployment error occurred Secret "libvirt-secret" not found  (non ceph job)
```

### DIFF
--- a/lib/dataplane/kustomization.yaml
+++ b/lib/dataplane/kustomization.yaml
@@ -2,6 +2,14 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
+secretGenerator:
+  - name: libvirt-secret
+    behavior: create
+    literals:
+      - LibvirtPassword=12345678
+    options:
+      disableNameSuffixHash: true
+
 resources:
   - dataplane-ssh-secret.yaml
   - nova-migration-ssh-secret.yaml


abays:
It seems like this `Secret` is needed for data plane deployment and not control plane deployment:

```
$ oc get osdpd -n openstack
NAME                        NODESETS             STATUS   MESSAGE
edpm-deployment-post-ceph   ["openstack-edpm"]   False    Deployment error occurred nodeSet: openstack-edpm error: Secret "libvirt-secret" not found
```

Perhaps we should add it to `lib/dataplane` instead?

psathyan:
> It seems like this `Secret` is needed for data plane deployment and not control plane deployment:
> 
> ```
> $ oc get osdpd -n openstack
> NAME                        NODESETS             STATUS   MESSAGE
> edpm-deployment-post-ceph   ["openstack-edpm"]   False    Deployment error occurred nodeSet: openstack-edpm error: Secret "libvirt-secret" not found
> ```
> 
> Perhaps we should add it to `lib/dataplane` instead?

You are right.

The control plane is called once but `data-plane` can be called multiple times. Since we are maintaining the same literal, I believe we should be good.

abays (approved):
Tested with VA1

/lgtm

fultonj:
It would be nice in a follow up if `12345678` could be a values.yaml or similar.

Merging to unblock va1.

PR #189, opened 2024-04-19
Add ssh-known-hosts service to dataplane

fultonj:
Nova requires that an SSH known hosts file is created on all hosts for live migration. All VAs and DTs require this service and we probably didn't notice it was missing because we're not testing live migration (but should). The recently added DTs already had this service.

Related:
https://github.com/openstack-k8s-operators/ci-framework/pull/1515


SeanMooney (approved):
yep this is requirefor nova-compute to start but also for live/cold migration.

without the ssh-known-host service the nova comptue conteinar will not be able to bind mount the know host file and it will fail to deploy.

PR #190, opened 2024-04-19
Remove duplicate secretGenerator

fultonj:
Redundant and without this change yamllint does not pass.


lib/dataplane/kustomization.yaml
```diff
@@ -2,14 +2,6 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 
-secretGenerator:
```
fultonj, line 5: see line 11

abays (approved):
/lgtm

PR #194, opened 2024-04-23
Split lib/dataplane into nodeset and deployment

abays:
Splits the `lib/dataplane` into sub-dirs for the `OpenStackDataPlaneNodeSet` and the `OpenStackDataPlaneDeployment`, so that they may be atomically utilizied.  Should _hopefully_ preserve backwards compatibility for existing VAs and DTs that want to render both by still using the base `lib/dataplane` dir.  I tested VA1 rendering with the existing logic and diffed it against the data plane templates generated by the new structure, and they were equal.


fao89 (approved):
thank you!

fultonj (approved):
/lgtm

PR #195, opened 2024-04-23
Split edpm-values into edpm-nodeset-values and edpm-deployment-values

fao89:


automation/vars/default.yaml
```diff
@@ -31,7 +31,7 @@ vas:
             osdpd edpm-deployment-pre-ceph --for condition=Ready
             --timeout=1500s
         values:
-          - name: edpm-values
+          - name: edpm-nodeset-values
             src_file: values.yaml
```
abays, line 35: This `values.yaml` was renamed I think

automation/vars/default.yaml
```diff
@@ -49,7 +49,7 @@ vas:
         values:
           - name: service-values
             src_file: service-values.yaml
-          - name: edpm-values-post-ceph
+          - name: edpm-nodeset-values-post-ceph
```
cjeanner, line 52: isn't it a deployment, meaning `edpm-deployment-values-post-ceph` instead?

fao89: actually there are some changes to nodeset, it add some ceph related variables:
https://github.com/fao89/architecture/blob/edpm_values/examples/va/hci/nodeset-values.yaml#L18-L23

cjeanner: hmmm in the framework, those data are set from the ceph.yml play/hook. but OK, let's keep it here. We don't edit it anyway from the framework afaik.

fao89:
Actually, this won't help,
I need to change the kustomization files to grab only nodeset/deployment and also have a kustomiztion that combines both for not breaking

PR #196, opened 2024-04-23
Allow lib/networking to be used without directly referencing sub-dirs

abays:
This change allows `lib/networking` to be used directly instead of having to explicitly point to each sud-dir, i.e.

```
components:
  - ../../lib/networking/metallb
  - ../../lib/networking/netconfig
  - ../../lib/networking/nad
```

can be instead written as:

```
components:
  - ../../lib/networking
```


leifmadsen (approved):
Makes sense to me. This does seem like a nice fixup to make it easier to load a single component vs multiple components.

PR #200, opened 2024-04-24
Remove extra nova filters

afazekas:
The default service filters supposed to be enough,
no need for these options.


afazekas:
Waiting for response from networking ..

rsafrono:
as the DifferentHostFilter now is enabled by default there are no objections from the networking team to switch to the defaults

PR #202, opened 2024-04-25
Enable glance sparse image upload

fultonj:
When Glance is configured with Ceph RBD there are performance benefits for storage usage and upload time when `rbd_thin_provisioning` is set.

https://specs.openstack.org/openstack/glance-specs/specs/victoria/approved/glance_store/handle-sparse-image.html

### DIFF
--- a/examples/va/hci/service-values.yaml
+++ b/examples/va/hci/service-values.yaml
@@ -37,6 +37,7 @@ data:
       store_description = "RBD backend"
       rbd_store_pool = images
       rbd_store_user = openstack
+      rbd_thin_provisioning = True
     glanceAPIs:
       default:
         replicas: 3


konan-abhi:
> False

It should be True, right?

examples/va/hci/service-values.yaml
```diff
@@ -37,6 +37,7 @@ data:
       store_description = "RBD backend"
       rbd_store_pool = images
       rbd_store_user = openstack
+      rbd_thin_provisioning = False
```
konan-abhi, line 40: the default value is False, if it is set to True then you will get performance benefits.

konan-abhi (approved):
Thank you!!

abays (approved):
/lgtm

fmount (approved):
/lgtm

PR #203, opened 2024-04-26
[example] Delete routes from NNCP for environment

leifmadsen:
Example PR to show how to drop generated objects not desired from the rendered object for specific environments vs removing the values.yaml entries, which are required by the base components.

Related to PR #177

### DIFF
--- a/examples/va/nfv/ovs-dpdk-sriov/nncp/kustomization.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/nncp/kustomization.yaml
@@ -22,3 +22,12 @@ components:
 
 resources:
   - values.yaml
+
+patches:
+  - target:
+      group: nmstate.io
+      version: v1
+      kind: NodeNetworkConfigurationPolicy
+    patch: |-
+      - op: remove
+        path: /spec/desiredState/routes
--- a/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
@@ -182,6 +182,12 @@ data:
         values:
           - 192.168.122.1
 
+  routes:
+    config:
+      - destination: 192.168.122.0/24
+        next-hop-address: 192.168.122.1
+        next-hop-interface: enp6s0
+
   rabbitmq:
     endpoint_annotations:
       metallb.universe.tf/address-pool: internalapi
--- a/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov/nncp/values.yaml
@@ -182,12 +182,6 @@ data:
         values:
           - 192.168.122.1
 
-  routes:
-    config:
-      - destination: 192.168.122.0/24
-        next-hop-address: 192.168.122.1
-        next-hop-interface: enp6s0
-
   rabbitmq:
     endpoint_annotations:
       metallb.universe.tf/address-pool: internalapi


leifmadsen:
@mnietoji I think this might be more of the solution you're looking for in PR #177

leifmadsen:
Before patch; resulting NNCP object:

```yaml
apiVersion: nmstate.io/v1
kind: NodeNetworkConfigurationPolicy
metadata:
  labels:
    osp/nncm-config-type: standard
  name: ostest-master-2
  namespace: openstack
spec:
  desiredState:
    dns-resolver:
      config:
        search: []
        server:
        - 192.168.122.1
    interfaces:
    - description: internalapi vlan interface
      ipv4:
        address:
        - ip: 172.17.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: internalapi
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "20"
    - description: storage vlan interface
      ipv4:
        address:
        - ip: 172.18.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: storage
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "21"
    - description: tenant vlan interface
      ipv4:
        address:
        - ip: 172.19.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: tenant
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "22"
    - description: ctlplane interface
      mtu: 1500
      name: enp6s0
      state: up
      type: ethernet
    - bridge:
        options:
          stp:
            enabled: false
        port:
        - name: enp6s0
          vlan: {}
      description: linux-bridge over ctlplane interface
      ipv4:
        address:
        - ip: 192.168.122.12
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: ospbr
      state: up
      type: linux-bridge
    routes:
      config:
      - destination: 192.168.122.0/24
        next-hop-address: 192.168.122.1
        next-hop-interface: ospbr
  nodeSelector:
    kubernetes.io/hostname: ostest-master-2
    node-role.kubernetes.io/worker: ""
```

After kustomize patch delete:

```yaml
apiVersion: nmstate.io/v1
kind: NodeNetworkConfigurationPolicy
metadata:
  labels:
    osp/nncm-config-type: standard
  name: ostest-master-2
  namespace: openstack
spec:
  desiredState:
    dns-resolver:
      config:
        search: []
        server:
        - 192.168.122.1
    interfaces:
    - description: internalapi vlan interface
      ipv4:
        address:
        - ip: 172.17.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: internalapi
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "20"
    - description: storage vlan interface
      ipv4:
        address:
        - ip: 172.18.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: storage
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "21"
    - description: tenant vlan interface
      ipv4:
        address:
        - ip: 172.19.0.7
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: tenant
      state: up
      type: vlan
      vlan:
        base-iface: enp6s0
        id: "22"
    - description: ctlplane interface
      mtu: 1500
      name: enp6s0
      state: up
      type: ethernet
    - bridge:
        options:
          stp:
            enabled: false
        port:
        - name: enp6s0
          vlan: {}
      description: linux-bridge over ctlplane interface
      ipv4:
        address:
        - ip: 192.168.122.12
          prefix-length: "24"
        dhcp: false
        enabled: true
      ipv6:
        enabled: false
      mtu: 1500
      name: ospbr
      state: up
      type: linux-bridge
  nodeSelector:
    kubernetes.io/hostname: ostest-master-2
    node-role.kubernetes.io/worker: ""
```

fultonj:
+2

leifmadsen:
> @leifmadsen thanks, yes dropping it like this should be ok. I am considering the full removal of ctlplane routes itself #208 as it shouldn't be needed afaik, it was originally added for baremetal use case and with discussion from @rabi it shouldn't be needed. Also did ran a test with uni-alpha job and that went fine.

Link to PR 208 doesn't resolve for me, so not sure what that is intended to link to. I guess the intent is to remove route control from the base components? If that's the case, then I suppose this isn't necessary? Is route control _ever_ necessary (not just for a particular use case, but in other potential use cases would it be necessary)?

I don't plan to merge this at all, as I opened it as an example of how to drop generated items by the base components that aren't necessary for overlays. If you feel this resolves the original issue, then we could close the other PR and move this one to review and have it tested.

For changes to validated architectures, we're looking for testing output to be provided on issues to identify that the PR works and doesn't result in regressions in other parts of the repository. For now that testing information will need to be provided manually as there is no automated functional testing for PRs currently. We're trying to avoid causing regressions within the repository.

leifmadsen:
OK I see PR #208 now and that at least seems reasonable to me. I haven't run it locally yet to see what the resulting manifests look like, but the approach seems sound. I'm going to close this example PR out for now, and we can reopen it if the approach in 208 doesn't seem to work out. Thanks!

PR #209, opened 2024-05-02
Generate replacement requirements documentation for components

leifmadsen:
Create a documentation generator that describes the replacements that are required for use of the component within a kustomization.yaml components list.


leifmadsen:
As this is more of a POC at this point, I'm going to probably stop picking at this in favour of obtaining feedback to the approach.

Reason for this PR, is that I felt the common components of the `lib/` directory could use some documentation to identify the expected parameters that need to be passed to the component to make it operational in an environment overlay. The replacements do not have default values, and errors will be present when running `kustomize build` if the components are included in the environment overlay. The generated documents provide the expected parameters to be passed via local-config ConfigMap.

A few things are missing:

* expectation that only a single ConfigMap name is used
  - [ ] create parameter hash per ConfigMap in case components use multiple configmaps
  - [ ] generated ConfigMap example would need to be done per-ConfigMap value
  - [ ] ConfigMap name needs to be updated

Other potential nice-to-haves:
* also show example of how to consume the component in an overlay
* use asciidoc so documentation could be used in product?
* include built docs in mkdocs

PR #210, opened 2024-05-02
Added libvirt-secret required for VA2

eshulman2:
Added libvirt-secret required for VA2, this should resolve an issue with edpm not being able to run the nova service.

Should resolve: https://issues.redhat.com/browse/OSPRH-6702

### DIFF
--- a/va/nfv/ovs-dpdk-sriov/edpm/kustomization.yaml
+++ b/va/nfv/ovs-dpdk-sriov/edpm/kustomization.yaml
@@ -24,6 +24,14 @@ resources:
   - baremetalset-password-secret.yaml
   - nova_ovs_dpdk_sriov.yaml
 
+secretGenerator:
+  - name: libvirt-secret
+    behavior: create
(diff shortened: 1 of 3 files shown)


va/nfv/ovs-dpdk-sriov/edpm/kustomization.yaml
```diff
@@ -24,6 +24,14 @@ resources:
   - baremetalset-password-secret.yaml
   - nova_ovs_dpdk_sriov.yaml
 
+secretGenerator:
```
Jaganathancse, line 27: This section is already covered:
https://github.com/openstack-k8s-operators/architecture/blob/main/lib/dataplane/nodeset/kustomization.yaml#L10-L16

I think this PR is not required.

eshulman2: It was missing when we tried to do some changes manually we should close this pr

PR #213, opened 2024-05-07
Move octavia-ca-passphrase to properly apply NamespaceTransformer

eduolivares:
For some reason, when ci-framework generates the
octavia-ca-passphrase Secret, it does not include any namespace.
The output obtained with `kustomize build` is different and the
openstack namespace is correctly included in this case.

With this patch, the octavia-ca-passphrase secret is moved to the
kustomization.yaml file where the NamespaceTransformer is defined too.
After this change, the openstack namespace is properly added to the
secret.

### DIFF
--- a/examples/dt/bgp/control-plane/kustomization.yaml
+++ b/examples/dt/bgp/control-plane/kustomization.yaml
@@ -2,13 +2,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-secretGenerator:
-  - name: octavia-ca-passphrase
-    literals:
-      - server-ca-passphrase=12345678
-    options:
-      disableNameSuffixHash: true
-
 components:
   - ../../../../dt/bgp/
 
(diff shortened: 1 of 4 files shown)


examples/dt/bgp/control-plane/kustomization.yaml
abays, line 1: The better approach is to move this from `example` dir [1] to the `dt` dir [2].  Then the `Transformer` [3] will dynamically add the namespace for you as it does for all other resources.

[1] https://github.com/openstack-k8s-operators/architecture/blob/a0c052be7e118bc237f3cda0aef602318e72e68e/examples/dt/bgp/control-plane/kustomization.yaml#L5-L10
[2] https://github.com/openstack-k8s-operators/architecture/blob/a0c052be7e118bc237f3cda0aef602318e72e68e/dt/bgp/kustomization.yaml#L25
[3] https://github.com/openstack-k8s-operators/architecture/blob/a0c052be7e118bc237f3cda0aef602318e72e68e/dt/bgp/kustomization.yaml#L5-L18

abays (approved):
/lgtm

PR #216, opened 2024-05-07
Adding label to track when architecture resources

fao89:


fao89:
my idea was to easily recognize where the resource came from, 
maybe this https://github.com/openstack-k8s-operators/install_yamls/pull/822 is enough to know if it came from install_yamls or here

leifmadsen:
> my idea was to easily recognize where the resource came from, maybe this [openstack-k8s-operators/install_yamls#822](https://github.com/openstack-k8s-operators/install_yamls/pull/822) is enough to know if it came from install_yamls or here

Well since this repo could eventually be used without install_yamls, or ci-framework etc as a production layer, I want to make sure we're not trying to integrate too heavily with the testing frameworks. I'm ok with having labels and such applied to the components to make it more clear how this is working, but I definitely don't want to embed this as a strong interface if it's not appropriate for non-testing framework deployments.

leifmadsen:
After some more thinking, I'm not sure if this is useful outside the context of the testing frameworks, and I'm not convinced it adds value to production deployments that might use the base components here for their environment overlays.

My opinion here is that we could review this in the future and determine if there is a common label set we want to apply, and then position that more centrally in a `common-labels` component or transformer etc. For now, my suggestion is to close this as unnecessary.

Looking for additional thoughts here as I could be convinced otherwise.

CC: @fultonj @abays

fao89:
I agree, I'm closing it

PR #221, opened 2024-05-09
Fix observability-operator name in readme

hjensas:
The `name` in the spec should be `cluster-observability-operator`.

Ref: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/openshift_obs/defaults/main.yml#L29C11-L29C40


psathyan (approved):

abays (approved):
/lgtm

PR #228, opened 2024-05-16
Temporary disable ovn_emit_need_to_frag in uni-alpha and zeta

karelyatin:
Due to the Related-Bug tests fails randomly, until the fix
is available in antelope temporary disabling the option.

Related-Bug: https://bugs.launchpad.net/neutron/+bug/2065701


karelyatin:
With the workaround applied issue is not seen, and this will help in getting real feedback from tests(non-network ones) instead of failing quite early due to this issues

psathyan (approved):

abays (approved):
/lgtm

PR #229, opened 2024-05-16
Fix octavia network attachment subnet range.

beagles:
Fixes the expected subnet IP range for the octavia network attachment. A previous change erroneously set this to the range for the tenant network instead of the provider network.

### DIFF
--- a/examples/dt/bgp/control-plane/nncp/values.yaml
+++ b/examples/dt/bgp/control-plane/nncp/values.yaml
@@ -189,9 +189,9 @@ data:
         "bridge": "octbr",
         "ipam": {
           "type": "whereabouts",
-          "range": "172.24.0.0/24",
-          "range_start": "172.24.0.30",
-          "range_end": "172.24.0.70"
+          "range": "172.23.0.0/24",
+          "range_start": "172.23.0.30",
+          "range_end": "172.23.0.70"
         }
       }
   external:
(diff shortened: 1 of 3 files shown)


karelyatin (commented):
/lgtm

@psathyan likely you pointing out this bit earlier over slack?

examples/dt/bgp/control-plane/nncp/values.yaml
```diff
@@ -189,9 +189,9 @@ data:
         "bridge": "octbr",
         "ipam": {
           "type": "whereabouts",
-          "range": "172.24.0.0/24",
-          "range_start": "172.24.0.30",
-          "range_end": "172.24.0.70"
+          "range": "172.23.0.0/24",
```
psathyan, line 192: Can we use user provider CIDR? If no, I would need to change in the downstream configurations that we pass.

psathyan: Thank you @beagles for the clarification.

psathyan (approved):

abays (approved):
/lgtm

abays:
Has this been tested?

psathyan:
> Has this been tested?

The `uni01alpha` downstream jobs have already picked up the change and have executed successfully in the last 2 runs.

PR #234, opened 2024-05-20
Move changeme statements

eshulman2:
Inline changeme statements within raw section were causing nove to fail in case they were not removed, moving statements out of line.


abays (approved):
/lgtm

PR #238, opened 2024-05-20
[manila] Adjust service replica counts

gouthampacha:
The default replica count we want for ManilaAPI and ManilaScheduler is 3; while we want ManilaShare to have 1 replica at most.

### DIFF
--- a/examples/dt/uni02beta/control-plane/service-values.yaml
+++ b/examples/dt/uni02beta/control-plane/service-values.yaml
@@ -44,9 +44,9 @@ data:
         [DEFAULT]
         debug = true
         enabled_share_protocols=cifs
-      replicas: 1
+      replicas: 3
     manilaScheduler:
-      replicas: 1
+      replicas: 3
     manilaShares:
       share1:
         networkAttachments:
--- a/examples/va/hci/service-values.yaml
+++ b/examples/va/hci/service-values.yaml
@@ -47,11 +47,15 @@ data:
   manila:
     enabled: true
     manilaAPI:
+      replicas: 3
       customServiceConfig: |
         [DEFAULT]
         enabled_share_protocols=nfs,cephfs
+    manilaScheduler:
+      replicas: 3
     manilaShares:
       share1:
+        replicas: 1
         customServiceConfig: |
           [DEFAULT]
           enabled_share_backends = cephfs
--- a/va/hci/edpm-post-ceph/nodeset/kustomization.yaml
+++ b/va/hci/edpm-post-ceph/nodeset/kustomization.yaml
@@ -124,6 +124,28 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
+      fieldPath: data.manila.manilaAPI.replicas
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.manila.template.manilaAPI.replicas
+        options:
+          create: true
+  - source:
+      kind: ConfigMap
+      name: service-values
+      fieldPath: data.manila.manilaScheduler.replicas
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.manila.template.manilaScheduler.replicas
+        options:
+          create: true
+  - source:
+      kind: ConfigMap
+      name: service-values
       fieldPath: data.manila.manilaShares.share1.customServiceConfig
     targets:
       - select:


fultonj:
Updating the service-values.yaml alone isn't sufficient since kustomize needs to pick those up and apply them. E.g. this PR worked:

https://github.com/openstack-k8s-operators/architecture/pull/112/files

Because we already had this:

https://github.com/openstack-k8s-operators/architecture/blob/main/va/hci/edpm-post-ceph/nodeset/kustomization.yaml#L69-L79

So you'd need something like the above. I'm doing a similar PR for cinderAPI and cinderBackup and I can share that if you need more detail.

fultonj:
I see this is updating not only va1 (HCI) but also Manila with NetApp

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni02beta/README.md

fultonj:
Here's an example PR for something similar for cinder https://github.com/openstack-k8s-operators/architecture/pull/239

gouthampacha:
> Updating the service-values.yaml alone isn't sufficient since kustomize needs to pick those up and apply them. E.g. this PR worked:
> 
> https://github.com/openstack-k8s-operators/architecture/pull/112/files
> 
> Because we already had this:
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/va/hci/edpm-post-ceph/nodeset/kustomization.yaml#L69-L79
> 
> So you'd need something like the above. I'm doing a similar PR for cinderAPI and cinderBackup and I can share that if you need more detail.

I didn't need it, @fultonj 




```
% git diff origin/main examples/va/hci/service-values.yaml
diff --git a/examples/va/hci/service-values.yaml b/examples/va/hci/service-values.yaml
index 6fe2c62..057373c 100644
--- a/examples/va/hci/service-values.yaml
+++ b/examples/va/hci/service-values.yaml
@@ -44,11 +44,15 @@ data:
   manila:
     enabled: true
     manilaAPI:
+      replicas: 3
       customServiceConfig: |
         [DEFAULT]
         enabled_share_protocols=nfs,cephfs
+    manilaScheduler:
+      replicas: 3
     manilaShares:
       share1:
+        replicas: 1
         customServiceConfig: |
           [DEFAULT]
           enabled_share_backends = cephfs
```

```
TARGET=/tmp/new
mkdir $TARGET
pushd architecture/examples/va/hci > /dev/null
kustomize build > $TARGET/nodeset-post-ceph.yaml
```

this gave me:
```
cat $TARGET/nodeset-post-ceph.yaml | grep manila: -A 41
  manila:
    apiOverride:
      route: {}
    enabled: true
    template:
      manilaAPI:
        customServiceConfig: |
          [DEFAULT]
          enabled_share_protocols=nfs,cephfs
        networkAttachments:
        - internalapi
        override:
          service:
            internal:
              metadata:
                annotations:
                  metallb.universe.tf/address-pool: internalapi
                  metallb.universe.tf/allow-shared-ip: internalapi
                  metallb.universe.tf/loadBalancerIPs: 172.17.0.80
              spec:
                type: LoadBalancer
        replicas: 3
      manilaScheduler:
        replicas: 3
      manilaShares:
        share1:
          customServiceConfig: |
            [DEFAULT]
            enabled_share_backends = cephfs
            enabled_share_protocols = cephfs
            [cephfs]
            driver_handles_share_servers = False
            share_backend_name = cephfs
            share_driver = manila.share.drivers.cephfs.driver.CephFSDriver
            cephfs_conf_path = /etc/ceph/ceph.conf
            cephfs_cluster_name = ceph
            cephfs_auth_id=openstack
            cephfs_volume_mode = 0755
            cephfs_protocol_helper_type = CEPHFS
          networkAttachments:
          - storage
          replicas: 1
```

gouthampacha:
> I see this is updating not only va1 (HCI) but also Manila with NetApp
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni02beta/README.md

yes, it was intentional.. 

```
% pushd architecture/examples/dt/uni02beta/control-plane > /dev/null
% kustomize build > $TARGET/unibetacontrolplane.yaml
% grep manila: $TARGET/unibetacontrolplane.yaml -A 42
  manila:
    apiOverride:
      route: {}
    enabled: false
    template:
      manilaAPI:
        customServiceConfig: |
          [DEFAULT]
          debug = true
          enabled_share_protocols=cifs
        networkAttachments:
        - internalapi
        override:
          service:
            internal:
              metadata:
                annotations:
                  metallb.universe.tf/address-pool: internalapi
                  metallb.universe.tf/allow-shared-ip: internalapi
                  metallb.universe.tf/loadBalancerIPs: 172.17.0.80
              spec:
                type: LoadBalancer
        replicas: 3
      manilaScheduler:
        replicas: 3
      manilaShares:
        share1:
          customServiceConfig: |
            [DEFAULT]
            debug = true
            enabled_share_backends=netapp
            [netapp]
            driver_handles_share_servers=False
            share_backend_name=netapp
            share_driver=manila.share.drivers.netapp.common.NetAppDriver
            netapp_storage_family=ontap_cluster
            netapp_transport_type=http
          customServiceConfigSecrets:
          - osp-secret-manila-netapp
          networkAttachments:
          - storage
          replicas: 1
```

lib/control-plane/openstackcontrolplane.yaml
```diff
@@ -111,9 +111,9 @@ spec:
       manilaAPI:
         networkAttachments:
           - internalapi
-        replicas: 1
+        replicas: 3
       manilaScheduler:
-        replicas: 1
+        replicas: 3
```
fmount, line 116: I don't like the fact that github does not allow me to comment on something that is not part of this change, but anyway: should L121 of this file be `replicas: 0` ?

fmount (changes requested):

fultonj:
> > Updating the service-values.yaml alone isn't sufficient since kustomize needs to pick those up and apply them. E.g. this PR worked:
> > https://github.com/openstack-k8s-operators/architecture/pull/112/files
> > Because we already had this:
> > https://github.com/openstack-k8s-operators/architecture/blob/main/va/hci/edpm-post-ceph/nodeset/kustomization.yaml#L69-L79
> > So you'd need something like the above. I'm doing a similar PR for cinderAPI and cinderBackup and I can share that if you need more detail.
> 
> I didn't need it, @fultonj

That's because this PR modified `lib/control-plane/openstackcontrolplane.yaml` so that every VA/DT will get 3 replicas if manila is enabled.

I removed  the change to `lib/control-plane/openstackcontrolplane.yaml` [1] and the changes to both service-values.yaml do not have any effect when I run `kustomize build`.

[1] 
```
[fultonj@runcible architecture{manila}]$ git diff
diff --git a/lib/control-plane/openstackcontrolplane.yaml b/lib/control-plane/openstackcontrolplane.yaml
index 5bea1a3..b336a78 100644
--- a/lib/control-plane/openstackcontrolplane.yaml
+++ b/lib/control-plane/openstackcontrolplane.yaml
@@ -111,9 +111,9 @@ spec:
       manilaAPI:
         networkAttachments:
           - internalapi
-        replicas: 3
+        replicas: 1
       manilaScheduler:
-        replicas: 3
+        replicas: 1
       manilaShares:
         share1:
           networkAttachments:
[fultonj@runcible architecture{manila}]$
```

lib/control-plane/openstackcontrolplane.yaml
fultonj: Per L109 `enabled: false` so the effect is the same.

If we set L121 to 0, then we should also set L114 and L116 to 0 as well and then ensure both service-values.yaml in this PR change update them to 3 1 1 respectively. We will also need additional lines in the kustomization files to actually set it.

fmount: ok, this explains why we have `replicas: 1` on a `share` that can't be up until the Ceph cluster (or the backend in general) is available.
My comment was mostly about the fact that you might have a topology where manila is enabled `enabled: true` from the very beginning, but, in case the backend (`Ceph` for instance) is deployed in a different stage you need to keep `replicas: 0` until the ctlplane is patched.
My understanding is that everything works because `Manila` is an optional service: it's still `enabled: false` at the ctlplane deployment stage, and we switch the boolean later in time. I don't think we should set `scheduler` and `api` to `0` because they can live w/o a `manilaShare` in any case, and the scope can be reduced to that particular share instance that would crash.
However, if such change generates a problem in other `kustomization.yaml` files that need to be patched as well, I'm ok to not change replicas and reduce the number of changes made by this patch.

leifmadsen:
This isn't required in other DTs and VAs?

fultonj:
> This isn't required in other DTs and VAs?

Not all DTs and VAs use manila. Only unibeta and hci (please correct me if I'm wrong @gouthampacha )

abays:
@fmount @gouthampacha Did we come to an agreement on this patch?

fmount:
> > I think this PR should:
> > 
> > * not change lib/control-plane/openstackcontrolplane.yaml
> > * change service-values.yaml of hci and uni02beta and thier kustomization files
> 
> I believe we are still waiting on the above to get done.

I don't know if I'm reading it wrong but looks like @gouthampacha addressed the requested changes (@fultonj correct me if I'm wrong).
I don't agree to leave `replica: 1` to `manilaShare/share1` in [1] because it should be on the VA/DT to update it according to the backend that is set (Ceph, netapp, etc), but I'm fine w/ the current approach if you think it's good.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L129

fultonj:
> > > I think this PR should:
> > > 
> > > * not change lib/control-plane/openstackcontrolplane.yaml
> > > * change service-values.yaml of hci and uni02beta and thier kustomization files
> > 
> > 
> > I believe we are still waiting on the above to get done.
> 
> I don't know if I'm reading it wrong but looks like @gouthampacha addressed the requested changes (@fultonj correct me if I'm wrong). 

Yes, you're right. Sorry I missed that. I rebased it and it looks good to me now.

> I don't agree to leave `replica: 1` to `manilaShare/share1` in [1] because it should be on the VA/DT to update it according to the backend that is set (Ceph, netapp, etc), but I'm fine w/ the current approach if you think it's good.

I basically agree with you on that. Not sure we should block this review on it though. In the default lib controlplane Manila is disabled. If you run `grep "enabled: false" openstackcontrolplane.yaml -B 1` you'll see we have that a lot of services disabled and with one replica. That in itself could be addressed for all of these services in another PR though.

Thus, I'd like to go ahead with merging this.

+2

> [1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L129

fmount:
> > > > I think this PR should:
> > > > 
> > > > * not change lib/control-plane/openstackcontrolplane.yaml
> > > > * change service-values.yaml of hci and uni02beta and thier kustomization files
> > > 
> > > 
> > > I believe we are still waiting on the above to get done.
> > 
> > 
> > I don't know if I'm reading it wrong but looks like @gouthampacha addressed the requested changes (@fultonj correct me if I'm wrong).
> 
> Yes, you're right. Sorry I missed that. I rebased it and it looks good to me now.
> 
> > I don't agree to leave `replica: 1` to `manilaShare/share1` in [1] because it should be on the VA/DT to update it according to the backend that is set (Ceph, netapp, etc), but I'm fine w/ the current approach if you think it's good.
> 
> I basically agree with you on that. Not sure we should block this review on it though. In the default lib controlplane Manila is disabled. If you run `grep "enabled: false" openstackcontrolplane.yaml -B 1` you'll see we have that a lot of services disabled and with one replica. That in itself could be addressed for all of these services in another PR though.
> 
> Thus, I'd like to go ahead with merging this.
> 
> +2
> 
> > [1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L129

Yeah, I didn't want to block the review, I'm ok to merge this and follow up if required.

fultonj (approved):
/lgtm
/approve

PR #241, opened 2024-05-21
Add missing next hop route to octavia network attachment

beagles:
This route is required to route the amphora controller openshift network to the load balancer management tenant network.


karelyatin (commented):
/lgtm

@beagles is this the fix for load balancers stuck in PENDING_CREATE?

Also seems need to have this fixed in ci-framework too so routes get generated correctly.

abays:
Has anyone tested this in a BGP DT deployment?

beagles:
> /lgtm
> 
> @beagles is this the fix for load balancers stuck in PENDING_CREATE?

It likely is. I've manually applied the changes to an environment that was being used for testing and amphorae were being created properly. 


> Also seems need to have this fixed in ci-framework too so routes get generated correctly.
 Ack.

fultonj:
@eduolivares @beagles is this still needed? 

If so, has it been tested recently so that it's safe to merge?

eduolivares:
> @eduolivares @beagles is this still needed?
> 
> If so, has it been tested recently so that it's safe to merge?

I haven't tested this patch, but maybe this is what would fix the problems we found in BGP jobs with octavia tests: "load balancers stuck in PENDING_CREATE".
I will try to test it tomorrow and will let you know.

weinimo:
lgtm (but I let you test it first)

eduolivares:
I have tested this patch on a BGP environment and it doesn't resolve the issue with "load balancers stuck in PENDING_CREATE"

fultonj:
> I have tested this patch on a BGP environment and it doesn't resolve the issue with "load balancers stuck in PENDING_CREATE"

OK, so it doesn't fix that particular issue.

It sounds like it didn't break the deployment either though, right?

If so should we merge it so it helps with adding the route that is required to route the amphora controller openshift network to the load balancer management tenant network?

@beagles WDYT?

beagles:
> > I have tested this patch on a BGP environment and it doesn't resolve the issue with "load balancers stuck in PENDING_CREATE"
> 
> OK, so it doesn't fix that particular issue.
> 
> It sounds like it didn't break the deployment either though, right?
> 
> If so should we merge it so it helps with adding the route that is required to route the amphora controller openshift network to the load balancer management tenant network?
> 
> @beagles WDYT?

@fultonj agreed!

abays:
This was tested with BGP DT, but not Unialpha (which it also changes).  Should we test that first before merging?

beagles:
> This was tested with BGP DT, but not Unialpha (which it also changes). Should we test that first before merging?

I've tested this myself via dt and octavia does complete deploying.
[zuul@controller-0 control-plane]$ oc get octavia
NAME      STATUS   MESSAGE
octavia   True     Setup complete

But I welcome someone else verifying :)

beagles:
Holding for the moment. Something may have changed recently that broke this deployment. Investigating.

beagles:
> Holding for the moment. Something may have changed recently that broke this deployment. Investigating.

We tried to include this in a testproject recently along with the ci-framework projects and it caused there to be problems early in the environment setup that appreared to be unrelated to this patch.  While I verified it manually by running the DT README instructions it is still required that we run a testproject . So I am setting up a test project that only includes the architecture PR to verify it is fine.

beagles:
Ran this with a dedicated test project and this patch doesn't interfere with successful deployment of the controlplane.

fultonj (approved):
/approve
/lgtm

PR #244, opened 2024-05-22
OSASINFRA DT

mandre:
A DT for deploying OpenShift on OpenStack, following the recommendations
from the OpenShift on OpenStack reference architecture [1], in
particular:
- Ceph for Image, Block Storage, and Object Storage (using RGW).
- Nova ephemeral using local storage
- Octavia + Amphora
- Manila + NFS

[1] https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.2/html-single/reference_architecture_for_deploying_red_hat_openshift_container_platform_on_red_hat_openstack_platform/index

### DIFF
--- a/examples/dt/osasinfra/dataplane-post-ceph.md
+++ b/examples/dt/osasinfra/dataplane-post-ceph.md
@@ -1,0 +1,77 @@
+# Configuring and deploying the post-Ceph dataplane
+
+## Assumptions
+
+- The pre-Ceph [dataplane](dataplane-pre-ceph.md) was already deployed and Ceph was manually installed afterwards
+
+## Initialize post-Ceph
+
+Switch to the "openstack" namespace
+```
+oc project openstack
+```
+Change to the osasinfra directory
+```
+cd architecture/examples/dt/osasinfra
+```
+Edit the [values.yaml](values.yaml) and [service-values.yaml](service-values.yaml)
+files to suit your environment.
+```
+vi values.yaml
+vi service-values.yaml
+```
+The ceph sections of [values.yaml](values.yaml) should have values like this.
+```yaml
+data:
+    ceph:
+        conf: $CONF
+        keyring: $KEY
+
+```
+Where the values of the variables above can be retrieved by
+running the following commands on the Ceph cluster.
+```shell
+CONF=$(cat /etc/ceph/ceph.conf | base64 -w 0)
+KEY=$(cat /etc/ceph/ceph.client.openstack.keyring | base64 -w 0)
+```
+
+Generate the post-Ceph dataplane nodeset CR.
+```
+kustomize build > nodeset-post-ceph.yaml
+```
+Generate the post-Ceph dataplane deployment CR.
+```
+kustomize build deployment > deployment-post-ceph.yaml
+```
+
+## Create post-Ceph CRs
+
+Create the nodeset CR
+```
+oc apply -f nodeset-post-ceph.yaml
+```
+Wait for post-Ceph dataplane nodeset setup to finish
+```
+oc wait osdpns openstack-edpm --for condition=SetupReady --timeout=600s
+```
+Create the deployment CR
+```
+oc apply -f deployment-post-ceph.yaml
+```
+
+Wait for control plane to be available after updating
+```
+oc wait osctlplane controlplane --for condition=Ready --timeout=40m
+```
+
+Wait for post-Ceph dataplane deployment to finish
+```
+oc wait osdpd edpm-deployment-post-ceph --for condition=Ready --timeout=1200s
+```
+
+## Finalize Nova computes
+
+Ask Nova to discover all compute hosts
+```bash
+oc rsh nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
+```
--- a/examples/dt/osasinfra/deployment/kustomization.yaml
+++ b/examples/dt/osasinfra/deployment/kustomization.yaml
@@ -1,0 +1,13 @@
+# This is the kustomization for the FINAL step, edpm-post-ceph
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - ../../../../dt/osasinfra/edpm-post-ceph/deployment
+  # - https://github.com/openstack-k8s-operators/architecture/dt/osasinfra/edpm-post-ceph/?ref=main
+  ## It's possible to replace ../../../dt/osasinfra/edpm-post-ceph/ with a git checkout URL as per:
+  ## https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
+
+resources:
+  - values.yaml
--- a/examples/dt/osasinfra/edpm-pre-ceph/deployment/kustomization.yaml
+++ b/examples/dt/osasinfra/edpm-pre-ceph/deployment/kustomization.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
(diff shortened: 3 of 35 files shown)


rlobillo:
Hi @mandre 

I think it's better to rename from DT to VA as it's more accurate considering the aim of this configuration. What do you think?

abays:
> > Hi @mandre
> > I think it's better to rename from DT to VA as it's more accurate considering the aim of this configuration. What do you think?
> 
> As I currently understand it, a DT is a configuration of OpenStack to cover a particular use-case. I think that's the case here for OpenShift on OpenStack, but I might be mistaken.
> 
> @abays what do you reckon?

The idea for DTs is that they cover a particular approach to deployment but also at the same time try to do as much as possible as pertains to that particular approach, such that it can be used in CI/QE in jobs to test as much as possible in that context.  Whereas a VA is narrower in scope and something we would intend end-users to consume and modify for their environments.  Whichever fits this PR is what I would use.

fultonj:
@mandre can you look into the most recent zuul issue?

mandre:
> @mandre can you look into the most recent zuul issue?

Done. I had forgotten to run `./create-zuul-jobs.py`.

I'm moving this PR back to draft as I have some unanswered questions whether it should be a DT or a VA. I'll also squash the commits before merging. The PR itself is ready for review, but not yet ready for merging IMO.

mandre:
recheck

mandre:
@abays @fultonj We've decided to target a DT first, I've moved the PR back to "Ready for reviews" and squashed the commits. Please take a look.

fultonj (commented):
Would you please add a README similar to the following:

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/README.md


The README should tell the deployer what they'll get (I see you listed some of that in the commit message) and provides the steps in order that they should follow to create this by applying the k8s manifests in the right order. Feel free to create and link to other .md files as in the example above.

mandre:
> Would you please add a README similar to the following:
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/README.md
> 
> The README should tell the deployer what they'll get (I see you listed some of that in the commit message) and provides the steps in order that they should follow to create this by applying the k8s manifests in the right order. Feel free to create and link to other .md files as in the example above.

Done! I've copied the docs from the HCI VA (since this is what the OSASINFRA DT is based on) and adapted the path and timeouts.

examples/dt/osasinfra/dataplane-post-ceph.md
```diff
@@ -0,0 +1,77 @@
...
+        conf: $CONF
+        keyring: $KEY
+
+```
+Where the values of the variables above can be retrieved by
+running the following commands on the Ceph cluster.
+```shell
+CONF=$(cat /etc/ceph/ceph.conf | base64 -w 0)
+KEY=$(cat /etc/ceph/ceph.client.openstack.keyring | base64 -w 0)
+```
```
fultonj, line 36: customizing nova to use ceph is deliberately not included since nova ephemeral is using local disk. +1

dt/osasinfra/edpm-post-ceph/nodeset/nova_ceph.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: nova-custom-ceph
```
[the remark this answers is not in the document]
mandre: I had tried that originally and if I recall correctly it complained of a missing configmap. Perhaps I forgot to change the service name, so going to give it another try.

Thanks for the link, that helps. The one you pasted points to the control plane doc, I think the correct link is https://openstack-k8s-operators.github.io/openstack-operator/dataplane/#_composable_services.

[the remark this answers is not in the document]
mandre: It worked! I indeed probably forgot to change the service name last time.

I've pushed the requested change, here's [the diff](https://github.com/openstack-k8s-operators/architecture/compare/edb3a96ac35ca7b024d4cb0be0d020d2f8bb5b44..49745cf917766371cc16a85b1366d2d35a013a2c).

fultonj (commented):
I'm +2 to merge this. @abays are you good with it?

I've followed the new .md files included in this PR to genereate the resulting manifests they look reasonable to me based on the description. I compared them to the va/hci CRs and I see what I expect. It differs in the following respects as it should.

- it octavia and its supproting resrouces (e.g. network)
- it's dataplane-nodeset-pre-ceph.yaml is identifical to the one for va/hci
- it's nodeset-post-ceph.yaml has octavia manila with NFS intsead of cephfs and no nova-custom-ceph (as expected)
- I noticed one difference in bridge name `ocpbr` vs `ospbr` [1] but after grepping deeper I see that usage in both "hci" and "osasinfra" looks consistent in both so it's not a problem.

**It shouldn't affect existing jobs since all the changes here are isolated, thus I think it's safe to merge.**

[1]  partial diff of nodeset-post-ceph.yaml in hci vs osasinfra
```diff
@@ -454,7 +457,8 @@
       ovnController:
         networkAttachment: tenant
         nicMappings:
-          datacentre: ospbr
+          datacentre: ocpbr
+          octavia: octbr
       ovnDBCluster:
         ovndbcluster-nb:
           dbType: NB
```

examples/dt/osasinfra/deployment/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+# This is the kustomization for the FINAL step, edpm-post-ceph
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - ../../../../dt/osasinfra/edpm-post-ceph/deployment
+  # - https://github.com/openstack-k8s-operators/architecture/va/hci/edpm-post-ceph/?ref=main
+  ## It's possible to replace ../../../va/hci/edpm-post-ceph/ with a git checkout URL as per:
```
abays, line 9: This comment is referencing VA1 and should probably be updated for this DT

examples/dt/osasinfra/edpm-pre-ceph/deployment/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - ../../../../../dt/osasinfra/edpm-pre-ceph/deployment
+  # - https://github.com/openstack-k8s-operators/architecture/va/hci/edpm-pre-ceph/?ref=main
+  ## It's possible to replace ../../../../va/hci/edpm-pre-ceph/ with a git checkout URL as per:
```
abays, line 8: Ditto here on referencing the wrong thing

examples/dt/osasinfra/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+# This is the kustomization for the FINAL step, edpm-post-ceph
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - ../../../dt/osasinfra/edpm-post-ceph/nodeset
+  # - https://github.com/openstack-k8s-operators/architecture/va/hci/edpm-post-ceph/?ref=main
+  ## It's possible to replace ../../../va/hci/edpm-post-ceph/ with a git checkout URL as per:
```
abays, line 9: Reference update needed

examples/dt/osasinfra/deployment/kustomization.yaml
mandre: Indeed, I just pushed an update where I fixed all occurrences.

PR #251, opened 2024-05-29
[uni01alpha] Split edpm kustomize into nodeset and deployment

psathyan:
This change set aligns with the way the rest of uni*jobs are deploying the dataplane. That is, preparing the nodeset followed by the deployment.

[ ] Unit testing


zuul.d/validations.yaml
```diff
@@ -78,7 +78,9 @@
     - examples/dt/uni01alpha
     - examples/dt/uni01alpha/control-plane
     - examples/dt/uni01alpha/control-plane/nncp
+    - examples/dt/uni01alpha/edpm
```
abays, line 81: Should `examples/dt/uni01alpha/edpm/nodeset` be included here as well?

psathyan: I dont have `examples/dt/uni01alpha/edpm/nodeset`

psathyan: `create-zuul-jobs.py` creates and validates these files. I believe it is working correctly.

abays: Ack

abays (approved):
/approve
/lgtm

PR #252, opened 2024-05-29
Add edpmServiceType to nova-custom-sriov srv

jamepark4:
Add  `edpmServiceType: nova` to the nova-custom-sriov service for the sriov architecture scenario.

### DIFF
--- a/va/nfv/ovs-dpdk/edpm/nodeset/nova_ovs_dpdk.yaml
+++ b/va/nfv/ovs-dpdk/edpm/nodeset/nova_ovs_dpdk.yaml
@@ -12,6 +12,7 @@ metadata:
   name: nova-custom-ovsdpdk
 spec:
   label: nova-custom-ovsdpdk
+  edpmServiceType: nova
   configMaps:
     - ovs-dpdk-cpu-pinning-nova
   secrets:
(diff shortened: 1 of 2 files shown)


gibizer (approved):
make sense to me

abays:
I know this is targeting the SRIOV-related NFV VAs, but there's another NFV VA2 (for OVS-DPDK only) that seems like it would need the same thing [1].  Just wondering if we should adjust them all at the same time.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/va/nfv/ovs-dpdk/edpm/nodeset/nova_ovs_dpdk.yaml

jamepark4:
> I know this is targeting the SRIOV-related NFV VAs, but there's another NFV VA2 (for OVS-DPDK only) that seems like it would need the same thing [1]. Just wondering if we should adjust them all at the same time.
> 
> [1] https://github.com/openstack-k8s-operators/architecture/blob/main/va/nfv/ovs-dpdk/edpm/nodeset/nova_ovs_dpdk.yaml

Yup makes sense I'll update that as well.

abays (approved):
/lgtm
/approve

PR #254, opened 2024-05-30
Disable multipath in Glance for uni05epsilon

sdatko:
In job we observe the issue related to multipathd
despite that the service is configured on ocp nodes. It causes early failure in tests when cirros image is attempted to be uploaded to Glance (Cinder backend). We disable it for now to enable Tempest testing.

Jira: https://issues.redhat.com/browse/OSPRH-7393


rebtoor (approved):

abays (approved):
/approve
/lgtm

Akrog:
The real issue is not with multipathing, as mentioned in the Jira, so I wouldn't disable it.

The first problem was with Cinder and galera being deployed as multi-master. This issue is fixed with https://github.com/openstack-k8s-operators/architecture/pull/255

The next problem is with the time it takes to fail logging to the inaccessible IPv6 address of the storage array.
In other words, the issue is a bug, because even when it takes 2 minutes to do the attach the upload should not fail.

fultonj:
> The real issue is not with multipathing, as mentioned in the Jira, so I wouldn't disable it.
> 
> The first problem was with Cinder and galera being deployed as multi-master. This issue is fixed with #255
> 
> The next problem is with the time it takes to fail logging to the inaccessible IPv6 address of the storage array. In other words, the issue is a bug, because even when it takes 2 minutes to do the attach the upload should not fail.

My preference is to update this PR to 

> The real issue is not with multipathing, as mentioned in the Jira, so I wouldn't disable it.
> 
> The first problem was with Cinder and galera being deployed as multi-master. This issue is fixed with #255
> 
> The next problem is with the time it takes to fail logging to the inaccessible IPv6 address of the storage array. In other words, the issue is a bug, because even when it takes 2 minutes to do the attach the upload should not fail.

Maybe we could update this PR so it looks like this:
```
# TODO enable multipath when OSPRH-7393 is resolved
# cinder_use_multipath = true
```
and then update OSPRH-7393 to note that this patch should be reverted once it's resolved?

My motivation would be to, get green jobs but to know that they might be green because of known issues which need to be fixed.

However, I defer to the cinder team.

sdatko:
I updated the content to mention the Jira not only in commit message :+1:

fultonj:
Per a discussion w/ Gorka we can merge this. We will revert this however after [OSPRH-7393](https://issues.redhat.com//browse/OSPRH-7393) is resolved

fultonj (approved):
/lgtm
/approve

PR #259, opened 2024-06-02
allow enabling igmp snooping

eshulman2:
allow enabling igmp snooping in NFV va to allow igmp job in integration pipeline


abays:
Should we add the custom `OpenStackDataPlaneService` here [1] as a replacement as well?

[1] https://github.com/openstack-k8s-operators/architecture/blob/71c3e690875d207da2141550b0223a875db5dcbd/examples/va/nfv/ovs-dpdk-sriov/edpm/nodeset/values.yaml#L179

abays (approved):
/lgtm

PR #266, opened 2024-06-05
Set mysql_wsrep_sync_wait for Glance and global debug option

fmount:
Until we get the patch [1] in a new version of the glance-operator, this patch allows to get the right default for the DTs. This patch can be partially reverted once the operator gets the new code.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/548

### DIFF
--- a/examples/dt/bgp/control-plane/service-values.yaml
+++ b/examples/dt/bgp/control-plane/service-values.yaml
@@ -9,6 +9,7 @@ data:
   glance:
     customServiceConfig: |
       [DEFAULT]
+      debug=True
       enabled_backends = default_backend:swift
       [glance_store]
       default_backend = default_backend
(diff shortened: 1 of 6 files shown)


examples/dt/bgp/control-plane/service-values.yaml
```diff
@@ -19,6 +19,8 @@ data:
       swift_store_endpoint_type = internalURL
       swift_store_user = service:glance
       swift_store_key = {{ .ServicePassword }}
+      [database]
```
konan-abhi, line 22: don't we need debug = True here before line 12?

fmount: yep thank you!

abays (approved):
/lgtm
/approve

konan-abhi (approved):

PR #267, opened 2024-06-05
Add workarond for nova to support  galera multimaster

gibizer:
This WA merged in nova-operator https://github.com/openstack-k8s-operators/nova-operator/pull/776
but decided not to be backported to beta. Still beta needs the WA to
work so we are adding the WA here. These configs are used by the
downstream integration job. Note that even the downstream beta
integration job uses the architecture from the main branch.


gibizer:
abandoned in favor of https://github.com/openstack-k8s-operators/architecture/pull/268

PR #270, opened 2024-06-06
Disable multipath in Nova for uni05epsilon

gibizer:
IPv6 and NetApp in the epsilon job does not work well with multipath
enabled. It leads to repeated errors in the nova-compute log during
volume mount that leads to slow VM boot that leads to tempest timeouts.

This PR disables multipath in nova in the job until the linked Jira is
resolved.

Related: https://issues.redhat.com/browse/OSPRH-7393

### DIFF
--- a/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
+++ b/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
@@ -1,0 +1,31 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nova-custom-config
+data:
+  25-nova-custom.conf: |
+    # NOTE(gibi): We need to disable multipath as IPv6 is not fully
+    # working in this job with NetApp.
+    # Enable multipath when OSPRH-7393 is resolved.
+    [libvirt]
+    volume_use_multipath = False
+
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: nova-custom
+spec:
+  label: dataplane-deployment-nova-custom
+  configMaps:
+    - nova-custom-config
+  secrets:
+    - nova-cell1-compute-config
+    - nova-migration-ssh-key
+  playbook: osp.edpm.nova
+  caCerts: combined-ca-bundle
+  edpmServiceType: nova
+  containerImageFields:
+    - NovaComputeImage
+    - EdpmIscsidImage
--- a/dt/uni05epsilon/edpm/nodeset/kustomization.yaml
+++ b/dt/uni05epsilon/edpm/nodeset/kustomization.yaml
@@ -18,3 +18,7 @@ transformers:
 
 components:
   - ../../../../lib/dataplane/nodeset
+
+
+resources:
+  - nova_custom.yaml
--- a/examples/dt/uni05epsilon/values.yaml
+++ b/examples/dt/uni05epsilon/values.yaml
@@ -148,4 +148,4 @@ data:
       - ovn
       - neutron-metadata
       - libvirt
-      - nova
+      - nova-custom


gibizer:
/test unit
```
{  could not resolve pipeline image stream tag root: imagestreamtags.image.openshift.io "pipeline:root" not found}
```

dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: ConfigMap
+metadata:
+  name: nova-custom-config
+data:
+  25-nova-custom.conf: |
+    # NOTE(gibi): We need to disable multipath as IPv6 is not fully
+    # working in this job with NetApp.
+    # Enable multipath when OSPRH-7393 is resolved.
+    [libvirt]
+    volume_use_multipath = False
```
SeanMooney, line 12: form a nova prespective multipath is enabeld by default and tested in alpha so i think this is ok to disabel for now form a compute perspective.

we need to confirm if we actually have multiple paths in the alpha job to validate the cidner/storage backend part of that however in the context of this job i think this is fine to proceed with.

SeanMooney (approved):

dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: nova-custom
+spec:
+  label: dataplane-deployment-nova-custom
+  configMaps:
+    - nova-custom-config
```
SeanMooney, line 22: as a side note because you hae used this name
i belive cifmw_edpm_deploy_nova_compute_extra_config will also allow job to override the default config map you are specifying above

gibizer: I'm not sure as that requires nova-custom service to be used but this job explicitly used the default nova service in the nodeset

SeanMooney: right 

the ci framework generates

nova-extra-config

so if you use that name for the configmap instead
then we will be able to use cifmw_edpm_deploy_nova_compute_extra_config

in the ci job to modify this if needed in the future.

it thought you had used the same name but you have not.

SeanMooney: supporting that is not strictly required but if you do respin using the same name will make things more consistent across the jobs

SeanMooney: maybe some missing context the ci framework always creates the 

nova-extra-config configmap even if you don't specify
cifmw_edpm_deploy_nova_compute_extra_config
its just empty in that case.

so you don't actlly need to use nova-custom for this to work you just need a custom dataplaneservice for nova that references nova-extra-config to be able to customise vai cifmw_edpm_deploy_nova_compute_extra_config

gibizer: I don't want to create a dependency here to the ci-framework behavior as the architecture repo is expected to be used without ci-framework too.

SeanMooney: this would not create a depency im not suggesting you remove the default you  are creating

anyway we can keep the divreggance i just mentioned it because it was easy to make both ways fo customisting this work
we don't actually need to respine for that

gibizer: wouldn't ci-framework override the CM content if I use the same CM name?

fultonj:
I see this as a follow up to https://github.com/openstack-k8s-operators/architecture/pull/254 but for Nova.
It makes sense to temporary disable it to unblock other dependencies so I'm +2 on this.

abays (approved):
/lgtm
/approve

fultonj (approved):
/lgtm
/approve

PR #271, opened 2024-06-06
Set neutron MTU according to actual value

rsafrono:
When deployed on OCP with default MTU settings actual MTU is limited to 1400 while neutron default is 1500. Wrong MTU setting causes failures during some tests execution.

### DIFF
--- a/examples/dt/uni06zeta/control-plane/service-values.yaml
+++ b/examples/dt/uni06zeta/control-plane/service-values.yaml
@@ -116,6 +116,7 @@ data:
       type_drivers = geneve,vlan
       tenant_network_types = geneve,vlan
       extension_drivers = qos,port_security,dns_domain_ports
+      path_mtu = 1400
       [ml2_type_geneve]
       vni_ranges = 1:65536
       max_header_size = 38


examples/dt/uni06zeta/control-plane/service-values.yaml
```diff
@@ -116,6 +116,7 @@ data:
       type_drivers = geneve,vlan
       tenant_network_types = geneve,vlan
       extension_drivers = qos,port_security,dns_domain_ports
+      path_mtu = 1400
```
karelyatin, line 119: let's hold, i think we would need some routes in network attachments. Temporary can get this fix too if route fixes are not yet landed in ci-framework

karelyatin:
<< When deployed on OCP with default MTU settings actual MTU is limited to 1400 while neutron default is 1500. Wrong MTU setting causes failures during some tests execution.

@rsafrono just to have more context is this needed only for test_mtu_sized_frame test or also for any other test?

rsafrono:
@karelyatin Recently I've hit only issue with test_mtu_sized_frames. But IIRC we've seen in the past also issues with QoS bw_limit tests, where iperf was reporting zero traffic because of the MTU mismatch. Setting path MTU to 1400 allowed iperf to send traffic and made the QoS bw_limit tests passing in that case.

examples/dt/uni06zeta/control-plane/service-values.yaml
abays: @karelyatin Do we still need to hold on merging this?

karelyatin:
> @karelyatin Recently I've hit only issue with test_mtu_sized_frames. But IIRC we've seen in the past also issues with QoS bw_limit tests, where iperf was reporting zero traffic because of the MTU mismatch. Setting path MTU to 1400 allowed iperf to send traffic and made the QoS bw_limit tests passing in that case.

@rsafrono ack to go this route for now

examples/dt/uni06zeta/control-plane/service-values.yaml
karelyatin: @abays can we get this in

abays: Can we get an ack on whether this has been tested?

karelyatin: @abays full testing on zeta job not done, just the affected tests were tested in zeta environment with this config and worked fine, also we have component jobs running with this config already so should be fine.

abays (approved):
/lgtm
/approve

PR #272, opened 2024-06-06
explictly define nova cell templated instead of relying on defaulting

SeanMooney:
This change simply adds the cell0 and cell1 templates
to the openstackcontrolplane.yaml in the shared lib dir.

This change decouples the Architecture repo from the current defaulting
behaivor in the nova-operator.

### DIFF
--- a/lib/control-plane/openstackcontrolplane.yaml
+++ b/lib/control-plane/openstackcontrolplane.yaml
@@ -153,6 +153,21 @@ spec:
         replicas: 3
       schedulerServiceTemplate:
         replicas: 3
+      cellTemplates:
+        cell0:
+          cellDatabaseAccount: nova-cell0
+          cellDatabaseInstance: openstack
+          cellMessageBusInstance: rabbitmq
+          conductorServiceTemplate:
+            replicas: 1
+          hasAPIAccess: true
+        cell1:
+          cellDatabaseAccount: nova-cell1
+          cellDatabaseInstance: openstack-cell1
+          cellMessageBusInstance: rabbitmq-cell1
+          conductorServiceTemplate:
+            replicas: 1
+          hasAPIAccess: true
   octavia:
     enabled: false
     template:


gibizer (approved):
looks good

fultonj:
Is replicas 1 still correct here?

Still needs downstream testing. It's been 3 weeks.

SeanMooney:
who are you expecting to do the downstream testing?
i certenlay am not expect to be the one to do that, i fully expect the initial testing to be automate in this repos ci. any additional testing i expect to be don't by the reviewers.


1 is correct we can increase the number of conductors but we do not expect to need too at least not in ci.

we have already made this change in the openstack-operator by  the way
https://github.com/openstack-k8s-operators/openstack-operator/pull/829
so the non architrued jobs are already no longer depending on the defaulting as the
base samples are now updated.


any change to the nova operator behavior will happen post release at this point so this is not urgent

fultonj:
> who are you expecting to do the downstream testing? i certenlay am not expect to be the one to do that, i fully expect the initial testing to be automate in this repos ci. any additional testing i expect to be don't by the reviewers.

We have some CI but we are not gating for uni jobs. We don't want to break unijobs so we request testing:

  https://github.com/openstack-k8s-operators/architecture/blob/main/docs/contributing/pull-request-testing.md

The history of this PR shows testing which was done:

  https://github.com/openstack-k8s-operators/architecture/pull/305

> 1 is correct we can increase the number of conductors but we do not expect to need too at least not in ci.
> 
> we have already made this change in the openstack-operator by the way [openstack-k8s-operators/openstack-operator#829](https://github.com/openstack-k8s-operators/openstack-operator/pull/829) so the non architrued jobs are already no longer depending on the defaulting as the base samples are now updated.
> 
> any change to the nova operator behavior will happen post release at this point so this is not urgent

I suspect this PR will be safe and it's small so I'll see if I can include it in other testing which I am doing.

abays (commented):
I tested this against VA1 and it worked fine.  If we could rebase it to fix conflict, I think we can approve and merge.

SeanMooney:
sorry i have been workign on a now public cve for the last two weeks so this was not a priority
I'm waiting for some ci result so i can take a look briefly and see if its a simple rebase.
if its complex i wont be able to work on this until the cve is fully resolved.

SeanMooney:
the conflict was trivial so should be good now

fultonj (approved):
/approve
/lgtm

fultonj (approved):
/approve
/lgtm

PR #274, opened 2024-06-06
Remove mariadb from OpenStackControlPlane

fultonj:
We are using galera not mariadb so remove mariadb. mariadb is disabled already but there is no need to leave it in our CR.


fultonj:
I want @dciabrin to confirm we should remove this before we do

dciabrin:
+1, let's remove it, the galera CR can already be deployed with 1-node or 3-node, making mariadb redundant.

abays (approved):
/lgtm
/approve

PR #278, opened 2024-06-07
Glance is missing extraMounts on uni02beta

fmount:
`extraMounts` were not under template, hence ignored.


abays (approved):
/lgtm
/approve

PR #282, opened 2024-06-10
Enable ceph-nfs for uni04delta

fmount:
uni04delta DT has `native_cephfs` backend while it should be configured w/ `ceph_nfs`.

### DIFF
--- a/examples/dt/uni04delta/control-plane/service-values.yaml
+++ b/examples/dt/uni04delta/control-plane/service-values.yaml
@@ -54,18 +54,17 @@ data:
     enabled: true
     customServiceConfig: |
       [DEFAULT]
-      enabled_share_backends = cephfs
-      enabled_share_protocols = cephfs
+      enabled_share_backends = cephfsnfs
+      enabled_share_protocols = nfs
 
-      [cephfs]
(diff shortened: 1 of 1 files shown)


fmount:
@jirimacku FYI, this should be required to enable ganesha in the job run

fmount:
Adding DNM label until I get some feedback

fultonj (commented):
I'm +2 on this but waiting for feedback from Jiri as per Francesco.

examples/dt/uni04delta/control-plane/service-values.yaml
```diff
@@ -54,18 +54,17 @@ data:
     enabled: true
     customServiceConfig: |
       [DEFAULT]
-      enabled_share_backends = cephfs
-      enabled_share_protocols = cephfs
+      enabled_share_backends=cephfsnfs
```
gouthampacha, line 57:
```suggestion
      enabled_share_backends = cephfsnfs
```

gouthampacha (approved):
/lgtm 

thanks @fmount

fultonj (approved):
/approve
/lgtm

PR #283, opened 2024-06-10
Config uni06zeta glance with Cinder

fmount:
As per the uni06zeta topology we need to config Glance with Cinder as a backend instead of Swift.


fultonj (approved):
/approve
/lgtm

PR #285, opened 2024-06-11
[BGP_DT01] Enable IGMP Snooping

eduolivares:
This neutron feature needs to be enabled in order to test some multicast scenarios on BGP DT01.


abays (approved):
/lgtm
/approve

PR #286, opened 2024-06-11
Ensure certificates are installed before Ceph is deployed

fultonj:
In order for Ceph RGW to use HTTPS we need to make sure the `install-certs` service runs on EDPM nodes which host Ceph. This service must run before Ceph is configured by the human operator or ci-framework. As per ci-framework PR 1821 the `cifmw_cephadm` Ansible role will stat the files installed by the `install-certs` service in order to know if HTTP or HTTPS will be used for the RGW service that it configures.

Add `install-certs` service to to edpm-pre-ceph nodeset for uni04delta (the non-HCI ceph job). Prior to this patch only the edpm-nova nodeset was using the `install-certs` service.

Move `install-certs` service from edpm-post-ceph run to edpm-pre-ceph run for HCI VA.

Required for Jira: https://issues.redhat.com/browse/OSPRH-7245


karelyatin:
Let's test this first before merging, because i recall this will break the actual service deploy like nova/ovn etc by moving install-certs service. Also in pre- phase what dataplane service involved for certs setup?

fultonj:
> Let's test this first before merging, 

ok...

> because i recall this will break the actual service deploy like nova/ovn etc by moving install-certs service. 

As long as ovn/nova are still being configured after install-certs it shouldn't make a difference to them. 

The change to `examples/va/hci/edpm-pre-ceph/nodeset/values.yaml` and `examples/va/hci/values.yaml` both produce the same list when combined.
```
>>> [1,2,3] + [4,5,6]
[1, 2, 3, 4, 5, 6]
>>> [1,2,3,4] + [5,6]
[1, 2, 3, 4, 5, 6]
>>> 
```
If service 5 needs service 4 to run before it, then it's still run before it and its deliverable is unchanged. The service merely copies the certificates to a directory. 

> Also in pre- phase what dataplane service involved for certs setup?

in pre-phase we're just going to copy the certs to the EDPM nodes so that when the ceph playbook run happens it has them. The only reason the two lists are split is so that ceph can run. I was going to not have a split in the list but we decided to stop deploying ceph with openstack tools.

fultonj:
Thanks @vakwetu and @karelyatin for explaining that `the dataplane goes through the list of enabled services to determine which certs to create`.

Thus, for HCI I'll need to not move but instead add a second calling of the install-certs service.

fultonj:
The intention is to have .crt and .key files in known paths on EDPM nodes so that the Ceph role [1] will use them when the following variables are passed:
```
cifmw_cephadm_certificate: "/etc/pki/tls/example.com.crt"
cifmw_cephadm_key: "/etc/pki/tls/example.com.key"
```
The certificates above are for haproxy to be configured with a VIP as described in [2]. This instance of haproxy runs on the EDPM nodes and is registered with keystone via:
```
openstack endpoint create --region regionOne {{ service_swift_uuid.stdout }} \
    public https://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
```
Merely including the `install-certs` service is not sufficient to create the crt and key files though.

The dataplane operator looks at the services to know if it has to create the certificate. None of the dataplane services in the pre-edpm-ceph list have certs configured right now. I was hoping to update the `ceph-hci-pre` service [3] to change this but the cert generation feature in dataplane-operator isn't currently designed for the simple case of issuing a cert with a fixed subject or uri which is the same for every node.

As a workaround if the required certificate was created out of band and stored as a k8s secret, then that secret could be included in a custom version of the `ceph-hci-pre` service [3] which mounts the secret like [4] and then the edpm-ansible's `edpm_ceph_hci_pre` role [4] would need to copy that certificate into place. If I go that route then this patch would not be necessary.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/53910f55b514c77e9eef2e8f2df0516f27b32d68/roles/cifmw_cephadm/README.md?plain=1#L60-L68

[2] https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/7/html-single/object_gateway_guide/index#configuring-high-availability-for-the-ceph-object-gateway_rgw

[3] https://openstack-k8s-operators.github.io/dataplane-operator/#_composable_services

[4] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml#L11

[5] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ceph_hci_pre/tasks/main.yml

fultonj:
Abandoning this PR in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/1865

PR #287, opened 2024-06-12
Update timeouts for uni-jobs

sdatko:
There are timeouts without any additional errors I observe on some CI nodes, probably with recent updates the deployment of dataplane takes slightly longer with LVM operator and newer OpenShift versions. This commit increases the timeouts and unifies them for all uni-jobs.

### DIFF
--- a/automation/vars/default.yaml
+++ b/automation/vars/default.yaml
@@ -8,7 +8,7 @@ vas:
             oc -n openstack wait nncp
             -l osp/nncm-config-type=standard
             --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured
-            --timeout=60s
+            --timeout=5m
         values:
           - name: network-values
             src_file: values.yaml
@@ -18,7 +18,7 @@ vas:
         wait_conditions:
           - >-
             oc -n openstack wait osctlplane controlplane --for condition=Ready
-            --timeout=30m
+            --timeout=60m
         values:
           - name: network-values
             src_file: nncp/values.yaml
@@ -29,7 +29,7 @@ vas:
           - >-
             oc -n openstack wait
             osdpns openstack-edpm --for condition=SetupReady
-            --timeout=600s
+            --timeout=10m
         values:
           - name: edpm-nodeset-values
             src_file: values.yaml
@@ -40,7 +40,7 @@ vas:
           - >-
             oc -n openstack wait
             osdpns openstack-edpm --for condition=Ready
-            --timeout=1500s
+            --timeout=30m
         values:
           - name: edpm-deployment-values
             src_file: values.yaml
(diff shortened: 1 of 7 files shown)


fultonj:
I'm supportive of this change in general for when you take it out of draft and rebase.

karelyatin:
Would be good to understand what exactly started taking more time as these jobs used to run always on same set of hardwares, does that changed? Just wanted to avoid any product regressions here, is it happening to all jobs or some specific job and specific hardware? or if it's about dev systems then it's something different and ok to do

sdatko:
> Would be good to understand what exactly started taking more time as these jobs used to run always on same set of hardwares, does that changed? Just wanted to avoid any product regressions here, is it happening to all jobs or some specific job and specific hardware? or if it's about dev systems then it's something different and ok to do

So, TL;DR I attempted to test the uni-jobs deployments with enabled LVM-operator and I was hitting control plane deployment timeouts. I increased the timeouts and unified them for all the uni-jobs – with those, the deployments I tested went fine.

fultonj:
> > Would be good to understand what exactly started taking more time as these jobs used to run always on same set of hardwares, does that changed? Just wanted to avoid any product regressions here, is it happening to all jobs or some specific job and specific hardware? or if it's about dev systems then it's something different and ok to do
> 
> So, TL;DR I attempted to test the uni-jobs deployments with enabled LVM-operator and I was hitting control plane deployment timeouts. I increased the timeouts and unified them for all the uni-jobs – with those, the deployments I tested went fine.

We've been using  `ci_local_storage` [1] to implement our SC for all uni jobs but that SC implementation is not suitable for production. We need to these jobs to use LVMS as implemented by `ci_lvms_storage` [2] as per OSPRH-860. LVM-operator uses more resources from k8s though.

[1] https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_local_storage
[2] https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_lvms_storage

fultonj (approved):
/approve
/lgtm

karelyatin:
> > > Would be good to understand what exactly started taking more time as these jobs used to run always on same set of hardwares, does that changed? Just wanted to avoid any product regressions here, is it happening to all jobs or some specific job and specific hardware? or if it's about dev systems then it's something different and ok to do
> > 
> > 
> > So, TL;DR I attempted to test the uni-jobs deployments with enabled LVM-operator and I was hitting control plane deployment timeouts. I increased the timeouts and unified them for all the uni-jobs – with those, the deployments I tested went fine.
> 
> We've been using `ci_local_storage` [1] to implement our SC for all uni jobs but that SC implementation is not suitable for production. We need to switch these jobs to use LVMS as implemented by `ci_lvms_storage` [2] as per OSPRH-860. LVM-operator uses more resources from k8s though.
> 
> [1] https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_local_storage [2] https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_lvms_storage

Thanks for sharing so the cause is atleast known now.
But doesn't look good with that much increase in time, almost timeout is doubled now with this patch. Do we know why it takes that much time and if there is scope of any improvement/follow ups etcs?

fultonj (approved):
/lgtm
/approve

PR #288, opened 2024-06-13
[uni02beta] Drop NFS from Glance

sdatko:
Blocked by the same reason as described in:
https://issues.redhat.com/browse/OSPRH-7396
Disabling temporarily for enabling the deployment.


abays (approved):
/lgtm
/approve

PR #290, opened 2024-06-14
Enable Manila in uni02beta

fmount:
Manila is disabled in uni02beta because of the missing enabled parameter replacement. This patch is supposed to enable it.


abays (approved):
/lgtm
/approve

PR #292, opened 2024-06-14
Increase timeout for uni01alpha once more

sdatko:
While for other uni-jobs adding 15 minutes was enough, for the mighty uni01alpha it appears we need even more time added. In this change I am proposing another 30 minutes so we can have working deplyment and then use some profilers to analyze what is the slowest part there.


sdatko:
So, actually we identified that for Uni-Alpha there was a problem with deploying ironic, though it did not break the deployment – it got stuck with LVM-operator, just complaining about being unable to provision a volume: https://issues.redhat.com/browse/OSPRH-7688

PR #297, opened 2024-06-18
Revert "Cinder wait for DB writes on reads"

Akrog:
The openstack-k8s-operators/mariadb-operator#229 switched the mariadb-operator to deploy in Active/Passive mode per default, instead of multimaster.

So we should no longer set `mysql_wsrep_sync_wait`.

Related PR: https://github.com/openstack-k8s-operators/cinder-operator/pull/401

This reverts commit cf4f3f6efe8d6f26277df9ffca82cd13869a6aa0.


fultonj (approved):
/approve
/lgtm

PR #300, opened 2024-06-20
Revert mysql_wsrep_sync_wait config option

fmount:
This is not a clean revert: we are removing `mysql_wsrep_sync_wait` because of the `galera` fix, but we would like to keep `debug=True` for CI troubleshooting purposes.


abays (approved):
/lgtm
/approve

PR #304, opened 2024-06-21
Add tainted worker to unibeta

lmiccini:
In order to run disruptive tests we would need to have a worker that is in charge of running only pods created by the testoperator with the proper tolerations, so that simulating a failure would not impact the pod running the tests.
A similar change has been submitted to extend the cluster topology downstream.

### DIFF
--- a/automation/net-env/uni02beta.yaml
+++ b/automation/net-env/uni02beta.yaml
@@ -1,1 +1,776 @@
-uni01alpha.yaml
+---
+instances:
+  ceph-0:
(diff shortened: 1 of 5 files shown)


lmiccini:
recheck

abays:
Has this been tested?

lmiccini:
> Has this been tested?

yes, we tested this downstream (details in slack).

abays (approved):
/lgtm
/approve

PR #305, opened 2024-06-21
Bump service replicas to 3

gibizer:
As all the DTs and VAs are defining 3 OCP master/worker nodes we can
scale up the openstack service replicas to 3 form 1 by default.

* this make our redundancy pattern closer to 17
* this can speed up tempest execution

There are couple of open questions:
* do we want to bump the replicas on the services that are disabled by
  default? e.g. heat?
* do we want to be selective? e.g. not bumping all the cell conductors
  and vncproxy services to 3

Related: [OSPRH-5645](https://issues.redhat.com//browse/OSPRH-5645)

### DIFF
--- a/lib/control-plane/openstackcontrolplane.yaml
+++ b/lib/control-plane/openstackcontrolplane.yaml
@@ -11,9 +11,9 @@ spec:
       databaseInstance: openstack
       secret: osp-secret
       barbicanAPI:
-        replicas: 1
+        replicas: 3
       barbicanWorker:
-        replicas: 1
+        replicas: 3
       barbicanKeystoneListener:
         replicas: 1
   ceilometer:
@@ -26,6 +26,8 @@ spec:
     apiOverride:
       route: {"haproxy.router.openshift.io/timeout": "60s"}
     template:
+      cinderAPI:
+        replicas: 3
       customServiceConfig: |
         # Debug logs by default, jobs can override as needed.
         [DEFAULT]
@@ -110,6 +112,7 @@ spec:
     template:
       databaseInstance: openstack
       secret: osp-secret
+      replicas: 3
   manila:
     apiOverride:
       route: {"haproxy.router.openshift.io/timeout": "60s"}
@@ -138,11 +141,18 @@ spec:
       networkAttachments:
         - internalapi
       secret: osp-secret
+      replicas: 3
   nova:
     apiOverride:
       route: {}
     template:
       secret: osp-secret
+      apiServiceTemplate:
+        replicas: 3
+      metadataServiceTemplate:
+        replicas: 3
+      schedulerServiceTemplate:
+        replicas: 3
   octavia:
     enabled: false
     template:
@@ -164,16 +174,25 @@ spec:
           dbType: NB
           networkAttachment: internalapi
           storageRequest: 10G
+          replicas: 3
         ovndbcluster-sb:
           dbType: SB
           networkAttachment: internalapi
           storageRequest: 10G
+          replicas: 3
+      ovnNorthd:
+        logLevel: info
+        nThreads: 1
+        replicas: 1
+        resources: {}
+        tls: {}
   placement:
     apiOverride:
       route: {}
     template:
       databaseInstance: openstack
       secret: osp-secret
+      replicas: 3
   rabbitmq:
     templates:
       rabbitmq:


tosky:
Please mark this as draft, it needs some acks from all maintainers of the services touched by this change.

lib/control-plane/openstackcontrolplane.yaml
```diff
@@ -35,7 +37,7 @@ spec:
           - storage
         replicas: 0
       cinderScheduler:
-        replicas: 1
+        replicas: 3
```
fmount, line 40: I think cinder-scheduler should be 1, but @Akrog to confirm

fmount (changes requested):
I left a comment about cinder-scheduler that should be reverted to 1, but I think 3 cinder APIs is ok.

lib/control-plane/openstackcontrolplane.yaml
```diff
@@ -164,16 +174,19 @@ spec:
           dbType: NB
           networkAttachment: internalapi
           storageRequest: 10G
+          replicas: 3
         ovndbcluster-sb:
           dbType: SB
           networkAttachment: internalapi
           storageRequest: 10G
+          replicas: 3
```
SeanMooney, line 182: of we are runing the ovn north and south db in a 3 replica cluster we should likely also do the same for OVNNorthd which is the deamon the process the db changes and responds to neutorn vai the websocket.

lib/control-plane/openstackcontrolplane.yaml
Akrog: I confirm, that's what we are recommending for production because it's going to solve a bunch of issues for deployments.
TBH for CI shouldn't matter, but I think it has value to test things following our recommendations.

SeanMooney: reading the document linked i agree keeping it at 1 for now is best.
with that said i think there is definitely an upstream conversation to be had about finally making cinder aware of placement and using it to track allocation against backends so that you can move form eveutally consists to a atomic allocations model to allow horizontal scaling

that will take time if the cinder community want to move in that direction so
for 18/antelope replica 1 make sense to me. this is not something we would expect to change in the lifetime of 18.

lib/control-plane/openstackcontrolplane.yaml
gibizer: bumped northd to 3

lib/control-plane/openstackcontrolplane.yaml
gibizer: set cinder-scheduler to

fultonj:
- cinder scheduler looks like it's been reverted to 1 so I think that's resolved
- I assume we're waiting for downstream testing before merging

gibizer:
> * cinder scheduler looks like it's been reverted to 1 so I think that's resolved
> * I assume we're waiting for downstream testing before merging

We need:
* feedback from the networking folks
* some downstream testing I will trigger

I will bring this patch up tomorrow on the podified call.

lib/control-plane/openstackcontrolplane.yaml
booxter: There is no relationship between `northd` and OVN RAFT members number, let me explain.

---

`northd` is active-passive HA. It grabs a OVS database lock, then connects to RAFT leaders (for NB and SB) and receives changes in NB and SB databases and translate between the two. The other replicas are just sitting there, waiting for the "main" `northd` to exit - at which point they will try to grab the lock and take over. In the meantime, each `northd` copy maintains its own copy of OVN IDL state in-memory - and connections to RAFT leaders - this has obvious load impact (memory, CPU - incl. for RAFT leaders that have to replicate changes to each "redundant" `northd` instance...)

I assume that, in general, `kubelet` should respawn the container with `northd` if it ever dies. It may take some time to respawn the process, but that's probably OK since `northd` works asynchronously to user visible operations, like Neutron API requests - or OVN database changes coming from Neutron agents or `ovn-controller`s.

Perhaps native `northd` active-passive HA implementation is more robust, in that some failure scenarios for `northd` may not be properly caught by `kubelet` and the container liveness checks. At the moment the check is just `pidof northd`, which is not ideal. (There's an `ovs-appctl status` command available for the service that would probably be a better way to confirm `northd` is not dead-locked.)

(The topic of inadequate liveness checks both in OCP and EDP containers was discussed before but nothing came out of it so far. I will report an issue to improve it in the next few days.)

Perhaps in some situations, having 2 `northd` instances may be a bit more robust, and HA takeover may happen quicker than container restart. (But see above: why do we care if NB changes are translated into SB in a second and not in 10 seconds **during disaster recovery / evacuation** that is an exceptional scenario?)

As I said, an argument **can** be made for bumping it to 2 (to keep the second passive replica). (Though I'd say we should have data to show the benefit is worth it.) But bumping to 3 is just excessive.

(Side comment: I don't think `neutron` ever talks to `northd` directly, via websocket or otherwise.)

Adding @otherwiseguy to keep me honest.

booxter:
Thanks for the PR.

I support OVN database replicas bump to 3. I don't think northd should be bumped; I am sure it should not be bumped to 3 / synchronized with the number of replicas for OVN databases.

gibizer:
> > * cinder scheduler looks like it's been reverted to 1 so I think that's resolved
> > * I assume we're waiting for downstream testing before merging
> 
> We need:
> 
> * feedback from the networking folks
> * some downstream testing I will trigger

I used https://gitlab.cee.redhat.com/ci-framework/ci-framework-testproject/-/merge_requests/375 to run unizeta with this PR. 
Here is the result https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/tripleo-ci-internal/build/a2b8189a50a042d0b97213517f81d323  As far as I see the tempest result isn't worse than the baseline we have in this job.(46 failed here, 55 failed in the integration pipeline.
However the tempest execution time is significantly sorter. (145minutes here, 234minutes in the integration pipeline)

> 
> I will bring this patch up tomorrow on the podified call.

gibizer:
> Thanks for the PR.
> 
> I support OVN database replicas bump to 3. I don't think northd should be bumped; I am sure it should not be bumped to 3 / synchronized with the number of replicas for OVN databases.

changed northd back to replicas:1

ccamposr:
> pport OVN database replicas bump to 3.

Component jobs tests with 1 replica and maybe it be interesting test with more than 1 if it is supported in integration. Even we could test HA cases with 1 and 2 and check if we detect some difference.

lib/control-plane/openstackcontrolplane.yaml
SeanMooney: ack i was not aware that it was internally looking an thought it could scale horizontally since that is not the case i would suggest keeping this at one or at most 2 for ha reason but we were originally dining this for performance reasons.
it sounds like this will not improve performance so if we want to do this for ha reasons we may want to drop this from the patch and consider that separately.

gibizer:
I run one more downstream uni zeta job on the last version of this PR. Tempest result and speed is the same as in the previous run. So this patch is good to go from my perspective.

fmount (approved):

fultonj (approved):
/lgtm
/approved

abays:
@SeanMooney Do you think we're ready to remove the hold now?

SeanMooney:
we have had review form networking storage and compute member so yes i think we can remove the hold based on that, @abays @fultonj if ye are happy with this feel free to drop the hold

fultonj (approved):
/lgtm
/approve

gibizer:
The tempest speedup due to this patch in downstream are summarized in https://redhat-internal.slack.com/archives/C04FMFMFFRC/p1719482652176879

PR #308, opened 2024-06-24
Add tainted worker to unibeta

lmiccini:
In order to run disruptive tests we would need to add one specially-tainted worker so that testoperator pods would be scheduled on top of it. 
This will isolate the test execution env from potential (eventual) OpenShift issues and would allow us to simulate things like "internet" api access, special routed network setups etc.

### DIFF
--- a/examples/dt/uni02beta/control-plane/nncp/kustomization.yaml
+++ b/examples/dt/uni02beta/control-plane/nncp/kustomization.yaml
@@ -21,3 +21,130 @@ components:
 
 resources:
   - values.yaml
+  - nodes.yaml
+
+replacements:
+  # Static Node IPs: node-3
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.node_3.internalapi_ip
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+          name: node-3
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=internalapi].ipv4.address.0.ip
+  - source:
+      kind: ConfigMap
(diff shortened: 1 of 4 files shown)


examples/dt/uni02beta/control-plane/nncp/kustomization.yaml
```diff
@@ -21,3 +21,130 @@ components:
...
+        path: /spec/nodeSelector/node-role.kubernetes.io~1worker
+  - target:
+      kind: NodeNetworkConfigurationPolicy
+      name: master-2
+    patch: |-
+      - op: add
+        path: /spec/nodeSelector/node-role.kubernetes.io~1master
+        value: ""
+      - op: remove
+        path: /spec/nodeSelector/node-role.kubernetes.io~1worker
```
sdatko, line 150: I wonder... I am not Kustomization expert, but could this maybe be moved to some lib, so when used by multiple DTs (potentially) we could reduce duplication?

abays: Technically this could be moved to `lib`, but I'm not sure what we would call it.  "Extra node"?  Anyhow, I think it could be addressed in a follow-up.

sdatko: At this point I also wonder if it should be in examples/dt/ or in dt/...

Anyway, I mainly want to avoid duplication if this will go to multiple DTs – or at least make it easier to handle.

For now it is fine as long as it it is just uni02beta

abays (approved):
/lgtm
/approve

PR #309, opened 2024-06-24
Don't use symlinks for automation/net-env/

hjensas:
The values used for ci coverage should match the actual DT it is covering.

NOTE: Only uni01alpha is updated to match the net-env of the actual scenario, the other files are still identical.

### DIFF
--- a/automation/net-env/uni01alpha.yaml
+++ b/automation/net-env/uni01alpha.yaml
@@ -1,414 +1,5 @@
 ---
 instances:
-  ceph-0:
-    hostname: ceph-0
-    name: ceph-0
-    networks:
-      ctlplane:
-        interface_name: eth1
-        ip_v4: 192.168.122.111
-        mac_addr: "52:54:00:3a:60:69"
-        mtu: 1500
-        netmask_v4: 255.255.255.0
-        network_name: ctlplane
-        prefix_length_v4: 24
-        skip_nm: false
-      external:
-        interface_name: eth1
-        ip_v4: 10.46.22.135
-        mac_addr: "52:54:00:3a:60:69"
-        mtu: 1500
-        netmask_v4: 255.255.255.192
-        network_name: external
-        prefix_length_v4: 26
-        skip_nm: false
-      internalapi:
-        interface_name: eth1.120
-        ip_v4: 172.17.0.111
-        mac_addr: "52:54:00:34:8e:48"
-        mtu: 1500
-        netmask_v4: 255.255.255.0
-        network_name: internalapi
-        parent_interface: eth1
-        prefix_length_v4: 24
-        skip_nm: false
-        vlan_id: 120
-      storage:
-        interface_name: eth1.121
-        ip_v4: 172.18.0.111
-        mac_addr: "52:54:00:10:9b:1d"
-        mtu: 1500
-        netmask_v4: 255.255.255.0
-        network_name: storage
-        parent_interface: eth1
-        prefix_length_v4: 24
-        skip_nm: false
-        vlan_id: 121
-      storagemgmt:
-        interface_name: eth1.123
-        ip_v4: 172.20.0.111
-        mac_addr: "52:54:00:0d:e9:95"
-        mtu: 1500
-        netmask_v4: 255.255.255.0
-        network_name: storagemgmt
-        parent_interface: eth1
-        prefix_length_v4: 24
-        skip_nm: false
-        vlan_id: 123
-  ceph-1:
-    hostname: ceph-1
-    name: ceph-1
-    networks:
-      ctlplane:
-        interface_name: eth1
-        ip_v4: 192.168.122.112
-        mac_addr: "52:54:00:f0:a5:5f"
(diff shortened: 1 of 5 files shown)


rebtoor:
/lgtm

Thanks @hjensas !

tosky (commented):
For the record, as the structure is changed, uni02beta, uni06zeta and uni07eta do not use ceph nodes

tosky:
I would like to call @sdatko, @queria , @jirimacku  and @arxcruz for another check on this

sdatko:
If I see correctly, this copies what was for uni01alpha for all existing uni jobs (instead of symlinks) and additionally it modifies the values for uni01alpha, yet not introduces similar changes for other DTs – as Luigi noticed, Beta, Zeta and Eta are not using Ceph, so IMVHO all those should be adjusted as well then? If I understand properly the idea...

So maybe it is worth to ask at this point – what is the general idea / problem this change attempts to resolve? Seem right now we just had unused elements in net-env for some cases, but it apparently wasn't a problem so far. So... what exactly changed?

Nevertheless, it would require verification for all the existing downstream jobs before merging.

cjeanner:
Hm, aren't those files used only for the `validate-architecture.yml` playbook context?

I recall seeing each DS job with its own networking-definition nested under 01-net-def.yml (or close name) in the ci-framework-jobs repository... ? @sdatko maybe I'm wrong there?

rebtoor:
> Hm, aren't those files used only for the `validate-architecture.yml` playbook context?
> 
> I recall seeing each DS job with its own networking-definition nested under 01-net-def.yml (or close name) in the ci-framework-jobs repository... ? @sdatko maybe I'm wrong there?

That's correct.

@sdatko @tosky 
Since i've added the concept of "net-env" files, i'd like to clarify better their usage here.

Each net-env file is basically a "mockup" needed to perform a "dry-run" of entire deployment, without applying anything, using the above mentioned [validate-architecture.yml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/architecture/validate-architecture.yml) playbook.
No actions will be performed in a real deployment so it doesn't have an impact on integration pipelines. 

So why are we needing these files? Because we need plausible values for each scenario despite we are running that playbook in a virtualized environment, without any compute/worker/networker/ceph nodes available (it will be time and resource consuming for "just" a validation) :) 

Keeping these files updated should be considered a best practice, but it's not strictly required.

Hope i clarified their usage better. Feel free to ping me if you need further info.

sdatko:
So, the net-env files are used only for tests, not during the deployments? The name could reflect that a little more.

So, what is the relation to `01-net-def.yaml` files from downstream ci-framework-jobs repo, because it seems a little redundant? Can the net-env be generated with some script easily maybe? Or maybe the net-def (and host-config, possibly?) shall be moved somewhere to architecture repo...?

hjensas:
The net-env files used here for test is what ends up after network_mapper run's. So essentially we should go and copy the content from each job and use that as the input. The issue I ran into here, is that when changeing the uni01alpha DT (adding a network), the validation job failed - I had to make changes to the net-env i.e adding the new network.

We can change the description of this, and say that it only updates uni01alpha to actaually match the scenrio. And then follow up with changes for the other DTs?

hjensas:
@sdatko, @queria , @jirimacku and @arxcruz - can we move this forward?

If you want all the mocks to actally reflect each scenario, feel free to push updates to this change.

abays:
> @sdatko, @queria , @jirimacku and @arxcruz - can we move this forward?
> 
> If you want all the mocks to actally reflect each scenario, feel free to push updates to this change.

Also wondering about this PR.  Any further input from @sdatko, @queria, @jirimacku and @arxcruz?

fultonj:
What was the original purpose of these symlinks?

hjensas:
> What was the original purpose of these symlinks?

I this @rebtoor is out sick, so I will try to answer.

I believe he added the mocks (as symlinks) for the test that does a form of validation between the content in architecture repo and the jinj2 templating in `ci_gen_kustomize_values`. Basically the files contain the data we have available in the `cifmw_networking_env_definition` variable in the `ci_gen_kustomize_values` templates.

Instead of adding a file (mock) for each DT/VA, one file with a combination of the data available in `cifmw_networking_env_definition` covering all the DTs/VAs was created - and symlinks was added to that file.

I think this can cause false positive results - if for example we reference data for a ceph node in the uni01alpha DT the validation would happily pass. But when we run the actual job, it would fail because uni01alpha does not have ceph. Similarly when we add network information for ironic or octavia to the mock, that would make it possible to get a false positive on DTs/VAs that don't have those.

To avoid potential false positive CI results I think these mocks should be updated to match the DT/VA it is validating. We can get that data easily from CI logs -  `/etc/ci/env/networking-environment-definition.yml` from the controller node should have the good data to use for the mock.

sdatko:
If I understand correctly it does not affect us, so... No blockers from uni-jobs perspective

fultonj (approved):
/approve
/lgtm

PR #310, opened 2024-06-25
epsilone: Do not force config drive

gibizer:
The nova default in 18 is to force the usage of the config drive over
the metadata-api. But we still want a dt where cloud-init is using
the metadata-api to initialize the guest to have test coverage.

Related: OSPRH-7901

### DIFF
--- a/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
+++ b/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
@@ -5,6 +5,10 @@ metadata:
   name: nova-custom-config
 data:
   25-nova-custom.conf: |
+    [DEFAULT]
+    # Override our defaults in this dt to get coverage for metadata-api based
+    # cloud-init scenarios
+    force_config_drive = False
     # NOTE(gibi): We need to disable multipath as IPv6 is not fully
     # working in this job with NetApp.
     # Enable multipath when OSPRH-7393 is resolved.


SeanMooney:
/lgtm

some context on this, neutron has internal logic called provisioning block that are used to wait for all port configuration to be done by the various agents before notifying nova that the port is "plugged" meaning the ovs config, DHCP and importantly metadata is configured.

when we use ovn neuton does not properly wait for metadata to be configured before notifying nova that the port is plugged.

as a result vms can boot before metadata is set up and cloud-init will fail to boot strap the VM with an sshkey or download userdata.

to mitigate this we use the force config drive flag in nova to ensure that all instance boot with a snapshot of the instance metadata in an attached config drive.

the epsilon job has been chose to disabel the forced config drive because this job also uses ipv6 so of all the jobs that is likely to have metadata issues this is the most likely and therefor will catch the most bugs.

separately from this we will also be disabling config drive in our downstream multi cell job as we will have per cell metadata deployed in that job.

SeanMooney (commented):
this is inline with what we discussed previously so I'm happy with this pr

tosky:
Uhm, just for the record, as far as I know uni-epsilon does not use IPv6

SeanMooney:
ah delta is the ipv6 jobs i was partly miss remembering we have ipv6 issue with the netap san which is why we are currently disabling multi path but your right epsilon is not using ipv6

it should be fine to test config drive disabled in any job the only reason to use a ipv6 jobs is we wont be testing ipv6 in the other downstream job we plan to turn config drive off in.

there is no really reason that it should depend on IP version anyway. so i think this is still fine to proceed with but we could do it to delta if we prefer.

karelyatin:
/lgtm

Thanks much @gibizer

karelyatin:
<< when we use ovn neuton does not properly wait for metadata to be configured before notifying nova that the port is plugged.
as a result vms can boot before metadata is set up and cloud-init will fail to boot strap the VM with an sshkey or download userdata.

@SeanMooney by chance do you have link for that issue? i was trying to find that few days back but couldn't.

abays (approved):
SMEs are good with this

/lgtm
/approve

SeanMooney:
> << when we use ovn neuton does not properly wait for metadata to be configured before notifying nova that the port is plugged. as a result vms can boot before metadata is set up and cloud-init will fail to boot strap the VM with an sshkey or download userdata.
> 
> @SeanMooney by chance do you have link for that issue? i was trying to find that few days back but couldn't.

i actually went looking for that myself upstream yesterday and could not find the bug report so i dug into the code a little. i think i know mostly what going on.

context is if your using an agent based backend like ml2/ovs or ml2/linuxbridge the metadata proxy is implemented by etiehr the DHCP agent or the l3 agent.
in both cases the metadata proxy is instancated in a network namespace when the DHCP server or router is created independently of creating a VM. additionally in the case of the DHCP agent we have a feature called provisioning blocks in neutron. how provisioning blocks works for ml2/ovs is that both the l2 agent and DHCP agent register as provisioning entities and both need to report back as complete before the neutron server actually notifies nova that plugging is complete 

https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/doc/source/contributor/internals/provisioning_blocks.rst#L4

the add_provisioning_component function is how we register entities that are required to report before provisioning is considered complete

ovn does support this and even adds an l2 entity

https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L737-L745
and it notifies as complete for both the DHCP entity and l2 
https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L746-L753
https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L1166-L1170

i don't see at a galnce where the ovn driver is adding the DHCP entity as a provisioning block by the way so that might also be a bug.

unlike the ml2/ovs implementation where the metadata proxy is set up when the DHCP port is bound or when the router namespace is created ovn delay that until the port show up and its done asynconously by the ovn metadata agent.

so i think the fix is either the DHCP entity needs to be added as a component or the likely better fix would be to add a metadata agent entity type and register that for ovn and set provision complete in the ovn metadata agent.

its possible this is related to the dhcp entity and a new metadata entity is not required.
the way the common code in the neturon server is it only waits for all provisioning components that are registered for a port so if we are not adding the dhcp component then its not waiting for _notify_dhcp_updated to set provisions complete before sending the notification to nova.

https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/neutron/plugins/ml2/drivers/ovn/mech_driver/mech_driver.py#L746-L753
https://github.com/openstack/neutron/blob/305e1451bbc980258d7f1accdbba83e5b08ab59e/neutron/db/provisioning_blocks.py#L109-L148

if i do a code search I'm not seeing it registered in the ovn mech driver at all but it might be done in a shared parent class.

https://github.com/search?q=repo%3Aopenstack%2Fneutron++provisioning_blocks.DHCP_ENTITY+path%3A%2F%5Eneutron%5C%2Ftests%5C%2Funit%5C%2Fplugins%5C%2Fml2%5C%2Fdrivers%5C%2Fovn%5C%2Fmech_driver%5C%2F%2F&type=code

I'm not convinced the dhcp issue is related to metadata but i think the lack of provsioning_block logic for the ovn metadata agent is more or less why we see this since we stop using the l3 and dhcp agents with ovn. that is the workaround deploy with the dhcp agent and enable metadata via the dhcp agent. we wont see this in an ovn + ironic scenario as a result.

PR #313, opened 2024-06-27
Always set glance replicas to 3

fmount:
rbac tests are failing and the root cause seems to have glance pods being restarted due to the large amount of requests coming from tempest. The same tests are not failing on unigamma and unidelta, where replica is set to 3.

### DIFF
--- a/examples/dt/uni07eta/control-plane/service-values.yaml
+++ b/examples/dt/uni07eta/control-plane/service-values.yaml
@@ -58,10 +58,8 @@ data:
     glanceAPIs:
       default:
         preserveJobs: false
-        replicas: 1
+        replicas: 3
         type: split
-    storageClass: ""
-    storageRequest: 1G
 
   swift:
     enabled: true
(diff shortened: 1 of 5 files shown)


konan-abhi (approved):

fultonj (approved):
/approve
/lgtm

examples/dt/uni07eta/control-plane/service-values.yaml
```diff
@@ -58,7 +58,7 @@ data:
     glanceAPIs:
       default:
         preserveJobs: false
-        replicas: 1
+        replicas: 3
         type: split
     storageClass: ""
     storageRequest: 1G
```
fmount, line 64: @fultonj should this need to be `10` instead of `1` ? or did we have `1` on purpose?

fultonj: We can use 10. I don't believe 1 was used for any specific reason.

konan-abhi (approved):

PR #314, opened 2024-06-27
Set UNI-Beta to replica: 1

fmount:
As per Jira [OSPRH-7396](https://issues.redhat.com//browse/OSPRH-7396) `extraMounts` have been removed from the current CR  via #288. For this reason increasing the replicas w/ a RWO PVC will cause issues.


abays (approved):
/lgtm
/approve

PR #327, opened 2024-07-08
Fix missing octavia-ca-passphrase secret

weinimo:
in uni06zeta.

Jira: [OSPRH-8311](https://issues.redhat.com//browse/OSPRH-8311)


abays:
This seems relatively safe to merge.  It just generates an extra `Secret`, so it shouldn't break things even if it doesn't end up fixing the problem.

abays (approved):
/lgtm
/approve

PR #328, opened 2024-07-11
Epsilon: Re-enable multipath

Akrog:
Previously we disabled multipath in Epsilon jobs, but now that we are using the webhook to change iSCSI login timeouts [1] they can be re-enabled.

As a side note, and for future reference, disabling multipathing should not have been in this architecture repo, because it's not an architecture specific thing, it's a deployment specific issue.

The right place to change this would be in the job specific repo.

[1]: https://github.com/openstack-k8s-operators/ci-framework/pull/2060

### DIFF
--- a/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
+++ b/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
@@ -9,11 +9,6 @@ data:
     # Override our defaults in this dt to get coverage for metadata-api based
     # cloud-init scenarios
     force_config_drive = False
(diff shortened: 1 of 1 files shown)


fultonj:
@Akrog @tosky has this been obsoleted by https://github.com/openstack-k8s-operators/architecture/pull/330 ?

Or do we still need this?

fultonj:
This proposed change has already been made by https://github.com/openstack-k8s-operators/architecture/pull/330 

Thus, I'm closing this PR. If I've misunderstood somehow, then please let me know.

PR #329, opened 2024-07-11
Increase osctlplane controlplane waiting time for Ready condition

rdiazcam:
The waiting time for OpenStack ControlPlane to be Ready needs to be increased from 600 to 900 seconds to avoids spurious timeouts since sometimes 10 minutes is not enough to reach the Ready condition.


abays (approved):
/lgtm
/approve

PR #330, opened 2024-07-15
Re-enable cinder multipath in glance

fmount:
This patch removes the workaround introduced for `OSPRH-7393`. The patches landed on both openstack, glance and cinder operators, hence we might want to test this change to verify the bug and re-enable cinder multipath (which is what customers do).

Jira: https://issues.redhat.com/browse/OSPRH-7415

### DIFF
--- a/examples/dt/uni05epsilon/control-plane/service-values.yaml
+++ b/examples/dt/uni05epsilon/control-plane/service-values.yaml
@@ -67,8 +67,7 @@ data:
       cinder_store_password = {{ .ServicePassword }}
       cinder_store_project_name = service
       cinder_catalog_info = volumev3::internalURL
-      # TODO: Enable multipath when OSPRH-7393 is resolved
-      # cinder_use_multipath = true
+      cinder_use_multipath = true
       [secondary]
       swift_store_create_container_on_put = True
       swift_store_auth_version = 3
--- a/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
+++ b/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml
@@ -9,11 +9,6 @@ data:
     # Override our defaults in this dt to get coverage for metadata-api based
     # cloud-init scenarios
     force_config_drive = False
-    # NOTE(gibi): We need to disable multipath as IPv6 is not fully
-    # working in this job with NetApp.
-    # Enable multipath when OSPRH-7393 is resolved.
-    [libvirt]
-    volume_use_multipath = False
 
 ---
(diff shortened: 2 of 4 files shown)


gibizer:
I think we also want to test the removal of the workaround from the nova side e.g. https://github.com/openstack-k8s-operators/architecture/blob/993a4a1250aa082a243abfbb4779f888f87ec69e/dt/uni05epsilon/edpm/nodeset/nova_custom.yaml#L14-L16 and similar.

fmount:
...  and also I think we're not ready to merge because we need an additional patch to land first.

fultonj:
I'm +2 on this, I'm just waiting for confirmation that it's been tested and then I'll remove the needs-testing label and merge it.

fultonj:
@sdatko @tosky I had a quick chat with @fmount and my understanding is that one of you will let me know when I can merge this and then we can update OSPRH-7415

examples/dt/uni05epsilon/control-plane/service-values.yaml
```diff
@@ -67,8 +67,6 @@ data:
       cinder_store_password = {{ .ServicePassword }}
       cinder_store_project_name = service
       cinder_catalog_info = volumev3::internalURL
-      # TODO: Enable multipath when OSPRH-7393 is resolved
-      # cinder_use_multipath = true
```
sdatko, line 71: Wait, isn't it wrong? I think default of this value is `false`, hence I had originally added `cinder_use_multipath = true` here – shoudn't this patch just remove the `# `? (Same for zeta and eta)

sdatko:
> @sdatko @tosky I had a quick chat with @fmount and my understanding is that one of you will let me know when I can merge this and then we can update [OSPRH-7415](https://issues.redhat.com//browse/OSPRH-7415)

We are running tests, but I tried with one other patch that was failing the deployment.
I will rerun only with this patch and let you know :+1:

examples/dt/uni05epsilon/control-plane/service-values.yaml
fmount: right, I somehow assumed that the default was true, but I'm wrong [1]!
Updating the patch!

[1] https://github.com/openstack/glance_store/blob/master/glance_store/_drivers/cinder/store.py#L386

sdatko:
Ok, I rekicked uni05epsilon to verify after recent update, around end of the day I will confirm the results

sdatko:
@abays Green light from our side :heart: :+1: :1st_place_medal:

abays (approved):
/lgtm
/approve

PR #332, opened 2024-07-15
uni-delta: really enable cinder-backup

tosky:
The default replica of cinder-backup is 0: like cinder-volumes, it requires a backend to be configured).

Both variants of uni-delta (base and IPv6) configure the service but don't set the replice, so it is not started.

Set replica 3 to test its HA mode (the old behavior in the last tripleo releases).

### DIFF
--- a/dt/uni04delta-ipv6/edpm/nodeset/kustomization.yaml
+++ b/dt/uni04delta-ipv6/edpm/nodeset/kustomization.yaml
@@ -46,6 +46,18 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
+      fieldPath: data.cinderBackup.replicas
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.cinder.template.cinderBackup.replicas
+        options:
+          create: true
+
+  - source:
+      kind: ConfigMap
+      name: service-values
       fieldPath: data.cinderVolumes.ceph
     targets:
       - select:
--- a/dt/uni04delta/edpm/nodeset/kustomization.yaml
+++ b/dt/uni04delta/edpm/nodeset/kustomization.yaml
@@ -46,6 +46,18 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
+      fieldPath: data.cinderBackup.replicas
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.cinder.template.cinderBackup.replicas
+        options:
+          create: true
+
+  - source:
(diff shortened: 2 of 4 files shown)


sdatko:
I am not Kustomize expert, but should not be there also entries in dt/uni04*/edpm/nodeset/kustomize.yaml similar to this one here: https://github.com/openstack-k8s-operators/architecture/blob/main/dt/uni05epsilon/kustomization.yaml#L56 ?

abays:
> I am not Kustomize expert, but should not be there also entries in dt/uni04*/edpm/nodeset/kustomize.yaml similar to this one here: https://github.com/openstack-k8s-operators/architecture/blob/main/dt/uni05epsilon/kustomization.yaml#L56 ?

Yes, something similar is needed because we don't have this field covered in `lib/control-plane/kustomization.yaml` [1].

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/kustomization.yaml

fultonj:
The most recent patch set correctly implements the replicas. I followed the steps here with this PR and the main branch:
https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni04delta/edpm.md#crs-nodeset
```
 1007  gh pr checkout 332
 ...
 1011  cd architecture/examples/dt/uni04delta
 1012  kustomize build > edpm-nodeset.yaml
 ...
 1016  git checkout main
 1018  kustomize build > edpm-nodeset2.yaml
```
You can see the generated file has what it needs.
```
[fultonj@runcible uni04delta{main}]$ diff -u edpm-nodeset.yaml edpm-nodeset2.yaml 
--- edpm-nodeset.yaml	2024-07-16 15:20:25.758417462 -0400
+++ edpm-nodeset2.yaml	2024-07-16 15:22:09.077012099 -0400
@@ -151,7 +151,7 @@
           backup_ceph_user = openstack
         networkAttachments:
         - storage
-        replicas: 3
+        replicas: 0
       cinderScheduler:
         replicas: 1
       cinderVolumes:
[fultonj@runcible uni04delta{main}]$ 
```

fultonj:
We effectively have the same code in hci (va1/gamma) and so we know the control plane can be deployed with 3 replicas of cinder backup when ceph is deployed. Thus, I think it's safe to merge this.

- https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/service-values.yaml#L13-L19

- https://github.com/openstack-k8s-operators/architecture/blob/main/va/hci/edpm-post-ceph/nodeset/kustomization.yaml#L69-L79

fultonj (approved):
/approve
/lgtm

PR #336, opened 2024-07-17
Workaround IPv6 DNS problem with hardcoded server

fultonj:
This is a quick and dirty hack to unblock IPv6 delta testing. This will be reverted and replaced with a variable containing a valid DNS server.

Jira: https://issues.redhat.com/browse/OSPRH-8771


fultonj:
Someone didn't require this workaround when testing w/ 4.16.3. If we can confirm we can address the problem by using this version in CI and then I'll close this PR.

fultonj:
This workaround is not necessary if you use OCP 4.16

PR #341, opened 2024-07-19
Revert "Disable Horizon service temporarily"

mcgonago:
This reverts commit 95e301f58fc066c3dc40a01426985f66dede49f6.

### DIFF
--- a/dt/uni04delta-ipv6/kustomization.yaml
+++ b/dt/uni04delta-ipv6/kustomization.yaml
@@ -194,15 +194,3 @@ replacements:
           - spec.nova.template.cellTemplates
         options:
           create: true
-
-  - source:
-      kind: ConfigMap
-      name: service-values
-      fieldPath: data.horizon.enabled
-    targets:
-      - select:
-          kind: OpenStackControlPlane
-        fieldPaths:
-          - spec.horizon.enabled
-        options:
-          create: true
--- a/examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
+++ b/examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
@@ -8,9 +8,6 @@ metadata:
 data:
   preserveJobs: false
 
-  horizon:
-    enabled: false
-
   cinderBackup:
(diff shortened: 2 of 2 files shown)


bshephar (approved):

abays:
Has anyone tested this revert in a `uni04delta-ipv6` downstream job?

dt/uni04delta-ipv6/kustomization.yaml
```diff
@@ -194,15 +194,3 @@ replacements:
           - spec.nova.template.cellTemplates
         options:
           create: true
-
-  - source:
```
fultonj, line 198: we should keep this

examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
```diff
@@ -8,9 +8,6 @@ metadata:
 data:
   preserveJobs: false
 
-  horizon:
-    enabled: false
```
fultonj, line 12: we just need to change this to true

fultonj:
Replaced by https://github.com/openstack-k8s-operators/architecture/pull/342

dt/uni04delta-ipv6/kustomization.yaml
sdatko: Imo we should not – in other uni jobs we do not need to specify that explicitly, it gets the default from lib/control-plane/openstackcontrolplane.yaml

examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
sdatko: (see https://github.com/openstack-k8s-operators/architecture/pull/341/files#r1687774153)

sdatko:
IMVHO This one is preferred over https://github.com/openstack-k8s-operators/architecture/pull/342

In yesterday build we tested that the Horizon deploys fine in IPv6 environment now.

fultonj (approved):
/lgtm
/approve

PR #342, opened 2024-07-22
Enable Horizon in delta-ipv6

fultonj:
Effectively everts commit 95e301f58fc066c3dc40a01426985f66dede49f6 while keeping option to change it in the kustomization file.

### DIFF
--- a/examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
+++ b/examples/dt/uni04delta-ipv6/control-plane/service-values.yaml
@@ -9,7 +9,7 @@ data:
   preserveJobs: false
 
   horizon:
-    enabled: false
+    enabled: true
 
   cinderBackup:
     customServiceConfig: |


fultonj:
@abays @sdatko is going to test this tonight and based on the outcome hopefully we can merge tomorrow.

sdatko:
So, I confirm that it worked and we had Horizon pod running...

However, in other uni-jobs we do not explicitly enable horizon and it gets deployed as well
– because of the defaults in [lib/control-plane/openstackcontrolplane.yaml](https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L94)

Hence, for consistency, I am more in favor of https://github.com/openstack-k8s-operators/architecture/pull/341 you closed yesterday. @fultonj

fultonj:
Going back to https://github.com/openstack-k8s-operators/architecture/pull/341 for consistency

PR #344, opened 2024-07-24
[uni02beta] Reenable file/NFS for glance

tosky:
Most of the issues described in https://issues.redhat.com/browse/OSPRH-7396 were addressed, at least on the controlplane side. As glance is mostly concerned with that part, it should work again.

This is a revert of
* b917c3129258d74bafe192454ee3f94097671b5e [uni02beta] Drop NFS from Glance
* 87d9931a9cf08be1b800fab6dd876d8a8ea5a191 Set UNI-Beta to replica: 1

### DIFF
--- a/dt/uni02beta/kustomization.yaml
+++ b/dt/uni02beta/kustomization.yaml
@@ -126,6 +126,17 @@ replacements:
           - spec.glance.template.glanceAPIs.default.type
         options:
           create: true
+  - source:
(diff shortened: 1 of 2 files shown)


tosky:
Tested internally - it works again (thanks to the uniqueness of the pod names for glance-api,  I guess)

abays (approved):
/lgtm
/approve

PR #346, opened 2024-07-30
DNM: test enable octavia in uni06zeta

lavraham:


fultonj:
This will only affect the unified job, not the component ci job.
Do we want to leave this false by default and enable it on a per job level instead?
Waiting to get feedback from @beagles @gthiemonge

lavraham:
> This will only affect the unified job, not the component ci job. Do we want to leave this false by default and enable it on a per job level instead? Waiting to get feedback from @beagles @gthiemonge
Octavia also part of uni-zeta job, I created this PR to enabled Octavia in the deployment.

PR #349, opened 2024-07-31
Disable extra mount in uni02beta

sdatko:
It fails again due to some NetApp issues (?).
In this change, we keep the configuration for it,
but just remove from kustomization the creation
of that extra mount, so it can be used as a workaround until the problem is resolved.

### DIFF
--- a/dt/uni02beta/kustomization.yaml
+++ b/dt/uni02beta/kustomization.yaml
@@ -81,6 +81,7 @@ replacements:
           - stringData.ontap-cinder-secrets\.conf
         options:
           create: true
+
   # Glance
   - source:
       kind: ConfigMap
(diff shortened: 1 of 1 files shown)


fultonj:
I'd prefer a removal to a comment out. But can't you just not override `data.glance.extraMounts` in the job definition if you don't want kustomize to affect it? I assume this workaround a problem, but isn't there a better way to do the workaround.

sdatko:
Hmm... let me see if it will not complain if I override it to empty list in the job...

sdatko:
Settomg `data.glance.extraMounts` to `[]` resulted in:
```
stderr: 'error: accumulating components: accumulateDirectory: "recursed accumulation of path ''/home/zuul/src/github.com/openstack-k8s-operators/architecture/dt/uni02beta'': fieldPath `data.glance.extraMounts` is missing for replacement source ConfigMap.[noVer].[noGrp]/service-values.[noNs]"'`
```

sdatko:
I know it is not super nice, but at the same time we are more interested to see tests results from other services in that scenario, rather that deployment failure again...

fultonj:
- merging this to unblock an important job with the understanding that it will be reverted soon

fultonj (approved):
/approve
/lgtm

fultonj:
The proposal to revert has arrived https://github.com/openstack-k8s-operators/architecture/pull/356

PR #352, opened 2024-08-01
[manila][update] Use CephFS NFS as backend for update job

lkuchlan:


tosky:
-1 : this is HCI VA, which has been established to have manila/cephfs, not an update job.
It can't be merged.

examples/va/hci/service-values.yaml
```diff
@@ -51,7 +51,7 @@ data:
       replicas: 3
       customServiceConfig: |
         [DEFAULT]
-        enabled_share_protocols=cephfs
+        enabled_share_protocols=nfs
```
fmount, line 54: -1: I understand this is about "updates", but this would result in breaking VA1 (unless there's something I missed in the entire updates context).
@lkuchlan not sure you can give me more context, but I think this won't work.
In addition, VA/HCI doesn't deploy ganesha (it's in the unidelta DT), so I'm not in favor of merging this change

examples/va/hci/service-values.yaml
```diff
@@ -68,7 +68,8 @@ data:
           cephfs_cluster_name = ceph
           cephfs_auth_id=openstack
           cephfs_volume_mode = 0755
-          cephfs_protocol_helper_type = CEPHFS
+          cephfs_nfs_cluster_id = cephfs
```
fmount, line 71: same concern expressed in my previous comment.

fmount (changes requested):

examples/va/hci/service-values.yaml
lkuchlan: Thanks, I thought it was an update job because of the ci-framework-baremetal-rhel9-va-hci-update-trunk-patches-base job. 
Where should the change be applied?

tosky: Maybe there a no jobs in ci-framework which deploys manila with ganesha, as there are no public jobs which use uni-delta.

gouthampacha: @tosky agreed

@lkuchlan think we should clone the base job into a downstream/component level update job and enable cephfs-nfs there instead, like we do for uni-delta.. i'd suggest doing this for completeness.

However, I think the base update job testing manila with Native CephFS is totally fine.. the point of updates testing would be to see the consequences of updating the OpenStack operators, and swapping out service container images... There shouldn't be a huge difference testing with CephFS-NFS since the Ceph cluster in the deployment isn't touched during the update process.

lkuchlan: So let's leave it as it is with Native CephFS. I'll just adjust the job to run the Manila Tempest tests.

PR #353, opened 2024-08-01
Deploy 2 replicas of DNS per default

stuggi:
like for glance, cinder, nova and all the other services we test with recommended number of replicas. That's why we should also test with DNS to have redundant number of replicas.

### DIFF
--- a/lib/control-plane/openstackcontrolplane.yaml
+++ b/lib/control-plane/openstackcontrolplane.yaml
@@ -45,7 +45,7 @@ spec:
   dns:
     template:
       options: []
-      replicas: 1
+      replicas: 2
   galera:
     enabled: true
     templates:


abays:
I tested this with CIFMW to deploy VA1 and everything succeeded.

@fultonj @sdatko Is this something you think we should have for every VA and DT?

fultonj:
> I tested this with CIFMW to deploy VA1 and everything succeeded.
> 
> @fultonj @sdatko Is this something you think we should have for every VA and DT?

It looks to me (`fgrep -R dns:`) like every VA/DT is going to get this except `examples/dt/uni04delta-ipv6/control-plane.yaml`.

I'd be surprised if this really caused a problem so I'm inclined to merge this. If it's a problem we can revert.

fultonj:
@stuggi can you give us a brief explanation on why you want this change?

I have no problem merging it but it would just be good to know if this is fixing a bug or just because we think there should be >1 for redundancy.

fultonj:
> > I tested this with CIFMW to deploy VA1 and everything succeeded.
> > @fultonj @sdatko Is this something you think we should have for every VA and DT?
> 
> It looks to me (`fgrep -R dns:`) like every VA/DT is going to get this except `examples/dt/uni04delta-ipv6/control-plane.yaml`.

I was mistaken. I can see that `uni04delta-ipv6` will also get two dns replicas. (I just had junk left over on my machine).

stuggi:
> @stuggi can you give us a brief explanation on why you want this change?
> 
> I have no problem merging it but it would just be good to know if this is fixing a bug or just because we think there should be >1 for redundancy.

If got the purpose of the VAs/DTs correct we want to test recommended number of replicas we suggest in docs for production, like for glance, cinder, nova and all the other services. That's why we should also test with DNS to have redundant number of replicas.

abays:
> > @stuggi can you give us a brief explanation on why you want this change?
> > I have no problem merging it but it would just be good to know if this is fixing a bug or just because we think there should be >1 for redundancy.
> 
> If got the purpose of the VAs/DTs correct we want to test recommended number of replicas we suggest in docs for production, like for glance, cinder, nova and all the other services. That's why we should also test with DNS to have redundant number of replicas.

I think this makes sense, especially if we're already using redundancy for other services.

fultonj (approved):
/approve
/lgtm

fultonj:
Does this need to be cherry picked to the branch `18.0.0-proposed` ?

PR #355, opened 2024-08-05
Enable ironic on delta-ipv6

afazekas:
The job currenly passing.
The ironic api test also can work after enabling.

According to examples/dt/uni04delta-ipv6/README.md 
The ironic service should be enabled here.


hjensas (approved):

abays (approved):
/lgtm
/approve

PR #356, opened 2024-08-05
Revert "Disable extra mount in uni02beta"

tosky:
It seems glance/NFS (or better, the combination of glance/NFS and
cinder/NFS) is stable enough, at least when used with different shares,
so that it can be reenabled.

This reverts commit 43ba5f8f471a0bf24774b2bc89ac3cb25c843ac8.


sdatko:
:-(

PR #359, opened 2024-08-06
Add document about cherry picking patches

fultonj:
Also, update DTs link in mkdocs.yml now that we have a lot of DTs.


abays (approved):
/lgtm
/approve

PR #360, opened 2024-08-06
Use pre/post Ceph servicesOverride in HCI deployments

fultonj:
Prior to this patch the OpenStackDataPlaneNodeSet changed before and after Ceph was deployed. This resulted in an incomplete nodeset services definition which resulted in incomplete updates in future deployments.

With this change the NodeSet will have the full services list but the pre-ceph and post-ceph deployments each pass a different servicesOverride list so the desired subset of services can be run before or after Ceph is deployed but the NodeSet services list remains complete.

Note that the NodeSet does not have the `nova-custom-ceph` service when it's initially created since that service is not defined until after Ceph is deployed. However, When `kustomize build` is run in `examples/va/hci` the full services list is defined in the NodeSet. Regardless, only a subset of services are run for the pre and post Ceph deployments.

Jira: https://issues.redhat.com/browse/OSPRH-9172

(cherry picked from 3ebad70fd9488ff2995553e8acc5e4fc85f24f1e)

### DIFF
--- a/examples/va/hci/dataplane-post-ceph.md
+++ b/examples/va/hci/dataplane-post-ceph.md
@@ -88,3 +88,12 @@ Ask Nova to discover all compute hosts
 ```bash
 oc rsh nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
 ```
+
+## Final OpenStackDataPlaneNodeSet services list
+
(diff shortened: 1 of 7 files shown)


abays (approved):
/lgtm
/approve

PR #361, opened 2024-08-06
Add telemetry to uni04delta-ipv6

vyzigold:


examples/dt/uni04delta-ipv6/control-plane.md
```diff
@@ -3,6 +3,40 @@
...
+  sourceNamespace: openshift-marketplace
+EOF
+
+# Apply the cr
+oc apply -f subscription.yaml
+
+# Wait for the deployment to be ready
+oc wait deployments/observability-operator --for condition=Available \
+    --timeout=300s
+```
```
abays, line 38: In theory this could be handled in this DT's Kustomize templates and could be added as a stage in the associated https://github.com/openstack-k8s-operators/architecture/blob/main/automation/vars/uni04delta-ipv6.yaml

fultonj: Is it automated elsewhere when you tested it downstream? 

I have no objection to this PR otherwise.

hjensas: ++ I would like to see that we try to consolidate on the automation here. 

For context, currently observability-operator is installed by `deploy-edpm.yml` which does `ansible.builtin.import_playbook: playbooks/02-infra.yml` - the `02-infra.yml` includes ci-framework role `openshift_obs` if `cifmw_deploy_obs: true`. AFICT only uni01alpha enabled the `cifmw_deploy_obs` option.

For this change to work, we would have to update the uni04delta-ipv6 job, toggling the same option.

I would rather see we implement it as suggested by Andrew here.

dt/uni04delta-ipv6/kustomization.yaml
```diff
@@ -92,3 +92,188 @@ replacements:
...
+
+  - source:
+      kind: ConfigMap
+      name: service-values
+      fieldPath: data.ironic.rpcTransport
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.ironic.template.rpcTransport
```
hjensas, line 104: I guess touching all these ironic values here is needed to avoid the OpenstackControlPlane update done to enable Ceph do not disable/change ironic config?

vyzigold: Hmm. I don't remember changing anything with ironic. Must be some leftover after a merge conflict and I somehow missed it when going through the PR :thinking: . Thank you for revealing this, I'll delete this and retest with the downstream testproject.

examples/dt/uni04delta-ipv6/control-plane.md
vyzigold: It is automated in the downstream job, but I don't mind moving it here.

vyzigold: I ended up including a yaml file containing the COO subscription and I added that into the control-plane's kustomization.yaml. I hope that works for you. I'll remove the line which deploys COO from the downstream MR and I'll do a recheck on the downstream testproject MR (both MRs depend on this PR) to test everything still works. I'll report back when I see it pass the CI.

examples/dt/uni04delta-ipv6/control-plane/kustomization.yaml
```diff
@@ -4,5 +4,6 @@ kind: Kustomization
 components:
   - ../../../../dt/uni04delta-ipv6
 resources:
+  - coo-sub.yaml
```
abays, line 7: This works to inject the `Subscription` for COO, because it will get processed here [1].  I do wonder though if it would be cleaner to have a separate stage for it before the control-plane one.  It's not a deal-breaker in my opinion, however.  Maybe @fultonj and @hjensas have other thoughts.

[1] https://github.com/openstack-k8s-operators/architecture/blob/6c766ab2086ad407a55d5daab7bf602bdb187655/automation/vars/uni04delta-ipv6.yaml#L17-L28

hjensas: I think a separate step would be better.

Also, I think we should move the coo-sub.yaml to the library, since this is something that multiple VA/DTs will use? Something in `lib/olm-deps`, and add the kustomize to apply in `examples/common` ?

Then the first automation stage for VA/DT's with Telemetry would be to apply `path: examples/common/coo-sub`? (I think that should work, but we may need to relax the assertions in ci-framework's `kustomize_deploy`[1] role to allow automation stages without values?)


[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/kustomize_deploy/tasks/execute_step.yml#L10-L11

abays: Those sound like good ideas to me

vyzigold: I like the idea too. I'll see what I can do.

vyzigold: Actually by moving the coo subscription to lib/olm-deps, it'll get picked up by examples/common/olm automatically. I assume that examples/common is used for each dt/va, so this way COO will be installed everywhere. I think that's OK to have it everywhere since it's listed as a dependency in the RHOSO docs and I've seen some efforts of enabling telemetry on all dts. Check it in the latest commit.

vyzigold: I rechecked my testproject downsteam which depends on these changes. It's struggling to deploy the dataplane for quite some time now (I think the hypervisor is running out of resources), but it got far enough to successfully deploy the control plane and I can confirm that this approach for deploying COO worked. I'll reach out once I manage to get a successful run of the testproject.

fultonj: @hjensas are you ok with this most recent update?

hjensas: Yes, I think it is fine. We will end up install COO always, but I saw the resource usage was discussed and agreed to be small enough to do it like this.

abays:
/hold

@vyzigold wants a @cjeanner review

vyzigold:
> /hold
> 
> @vyzigold wants a @cjeanner review

That's not required. I just saw that I received a LGTM and not an approval. The first message in the PR suggested to add him as a reviewer for approval, so that's what I did. Thanks for the approval. I'm really happy this is finally merged.

PR #363, opened 2024-08-08
Changing ntp to ci-framework default

afazekas:
This parameter is excepted to be overriden with
cifmw_ci_gen_kustomize_values_ntp_srv to your own
clock infra.


cjeanner:
/lgtm

this is preventing leaking downstream-only names, and following the best practices. Thanks for pointing the correct parameter name to use in the CI. We can add a linter to ensure this name isn't pushed in later PR.

abays:
Have this been tested?  If so, it looks good to me.

abays (approved):
/lgtm
/approve

PR #364, opened 2024-08-09
[NFV] Update the osdpns timeout for SetupReady

lewisdenny:
This patch bumps the timeout we use for the `OpenStackDataPlaneNodeSet` to 90m.

The bump is needed because the NFV deploys to real baremetal, 30m is enough for virtual baremetal but real machines take way longer to provision.

90m was chosen as we saw a failed job that took 70m to get to SetupReady

Jira: [OSPRH-9348](https://issues.redhat.com//browse/OSPRH-9348)

### DIFF
--- a/automation/vars/default.yaml
+++ b/automation/vars/default.yaml
@@ -204,7 +204,7 @@ vas:
           - >-
             oc -n openstack wait
             osdpns openstack-edpm --for condition=SetupReady
-            --timeout=60m
+            --timeout=90m
         values:
           - name: edpm-nodeset-values
             src_file: values.yaml


fultonj:
Please rebase to resolve conflicts.

saneax:
verified, this works

abays:
@lewisdenny Just curious: what sort of real baremetal hardware are you seeing these long times with?  I definitely agree that baremetal takes longer than virtual because of the costly reboots, but I've never seen provisioning itself take more than 30 minutes in any environment I've used.  The need for 90 minutes just sounds like something is not right with the environment.

pablintino:
> @lewisdenny Just curious: what sort of real baremetal hardware are you seeing these long times with? I definitely agree that baremetal takes longer than virtual because of the costly reboots, but I've never seen provisioning itself take more than 30 minutes in any environment I've used. The need for 90 minutes just sounds like something is not right with the environment.

@abays @lewisdenny Agree. I'm deploying NFV in our envs and it provisions in less than 30 mins, no matter if it's our new HW or the NFV QE HW. Something else must be happening underneath to make it take +30 mins.

lewisdenny:
Agreed, I have reached out to the user to understand why the node is taking so long to provision, we will troubleshoot it further but in the meantime they are `Depends-On`ing this patch to move forward. 

I'll just put this one on hold for now

/hold

lewisdenny:
No update yet from the original reporter of the issue so I will close this one, if it really needs to be bumped I'll reopen.

PR #365, opened 2024-08-11
increased NFV timeout

eshulman2:
imcreased nodeset and deployment timeout for NFV jobs due to repeating failures with wait command reaching its timeout


cjeanner:
/lgtm
Won't change the run time if it ends earlier than the set timeout, but would also avoid to re-kick the job when it times out with the previous values.

pojadhav (commented):
LGTM

abays (approved):
/lgtm
/approve

PR #368, opened 2024-08-14
Sync 18.0.0-proposed branch with current main branch

abays:
Requested by CI/QE for OSP 18


abays:
We're just going to delete `18.0.0-proposed` and recreate instead

PR #369, opened 2024-08-16
HCI+NFV DT

mnietoji:
Based on nfv ovs-dpdk-sriov and HCI VA
Due to some limitations on the available hardware, only 2 compute nodes are used while the recommended for ceph is 3. But this is ok for the purpose of the testing to be done on this environment


fultonj (approved):
/approve
/lgtm

PR #374, opened 2024-08-23
Enable Downstream trigger job in the architecture repo

raukadah:
In order to test VA HCI related changes, we used to run testproject in the downstream.

downstream-va-hci-trigger-job drops the need of running testproject for VA HCI related changes.

An developer can trigger the job by adding `trigger github-experimental` in the pull request comment section. The trigger job will create a review on
https://review.rdoproject.org/r/q/project:tripleo-downstream-trigger-nested-virt repo with proper depends-on with the origin pr and dependendent change.

From tripleo-downstream-trigger-nested-virt gerrit review, it will start VA HCI job downstream in rdo-check pipeline. Once the job finishes, It will post the results on the gerrit review. Devs can verify the results before merging it.

Note: It modifies create-zuul-jobs.py to include experimental jobs.

### DIFF
--- a/zuul.d/trigger-jobs.yaml
+++ b/zuul.d/trigger-jobs.yaml
@@ -1,0 +1,14 @@
+- job:
+    name: architecture-downstream-va-hci-trigger-job
+    parent: downstream-va-hci-trigger-job
+    description: |
+      This job will trigger a VA HCI job downstream.
+      We can invoke this job by adding following comment
+      `\trigger github-experimental` in the github pull request comment.
+      These jobs runs in github-experimental-trigger pipeline.
+    files:
+      - ^automation/vars/hci.yaml
+      - ^lib
+      - ^examples/common
+      - ^examples/dt/hci
+      - ^va/hci
--- a/create-zuul-jobs.py
+++ b/create-zuul-jobs.py
@@ -53,7 +53,8 @@ with open('./zuul.d/validations.yaml', 'w+') as zuul_jobs:
 _PROJECTS.sort()
 with open('./zuul.d/projects.yaml', 'w+') as zuul_projects:
     struct = [{'project': {'github-check': {'jobs': _PROJECTS},
-                           'github-gate': {'jobs': _PROJECTS}
+                           'github-gate': {'jobs': _PROJECTS},
+                           'github-experimental-trigger': {'jobs': ['architecture-downstream-va-hci-trigger-job']}
                            }
                }
               ]
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -18,5 +18,8 @@
       - rhoso-architecture-validate-uni05epsilon
       - rhoso-architecture-validate-uni06zeta
       - rhoso-architecture-validate-uni07eta
+    github-experimental-trigger:
+      jobs:
+      - architecture-downstream-va-hci-trigger-job
     github-gate:
       jobs: *id001


raukadah:
Here is [an example](https://github.com/openstack-k8s-operators/ci-framework/pull/2222#issuecomment-2304818571) of trigger job in action with [results](https://github.com/openstack-k8s-operators/ci-framework/pull/2222#issuecomment-2305965654).

fultonj:
I see the logs from the ceph deploy from the new trigger and it looks good. Thank you @raukadah this will be helpful. I assume we'll keep gating with `podified-multinode-hci-deployment-crc` but add this additional command when necessary to see downstream result when necessary. Do you agree?

I'm +2 to add this feature and don't see anything wrong with the patch myself.

https://sf.hosted.upshift.rdu2.redhat.com/logs/54/54154/5/check-rdo/ci-framework-baremetal-static-node-rhel9-va-hci/a774960/logs/controller-0/ci-framework-data/logs/ci_script_001_run_deploy.log

fultonj:
I do see a red result here `test-zuul`. 

https://github.com/openstack-k8s-operators/architecture/actions/runs/10518027471/job/29143170869?pr=374

raukadah:
> I see the logs from the ceph deploy from the new trigger and it looks good. Thank you @raukadah this will be helpful. I assume we'll keep gating with `podified-multinode-hci-deployment-crc` but add this additional command when necessary to see downstream result when necessary. Do you agree?
> 
> I'm +2 to add this feature and don't see anything wrong with the patch myself.
> 
> https://sf.hosted.upshift.rdu2.redhat.com/logs/54/54154/5/check-rdo/ci-framework-baremetal-static-node-rhel9-va-hci/a774960/logs/controller-0/ci-framework-data/logs/ci_script_001_run_deploy.log

@fultonj yup! The goal of this job to trigger when we need downstream results. In order to avoid the misuse, We are also working on [github-action pr ](https://github.com/openstack-k8s-operators/ci-framework/pull/2256) to restrict triggering to approvers only.

zuul.d/trigger-jobs.yaml
```diff
@@ -0,0 +1,13 @@
+- job:
+    name: architecture-downstream-va-hci-trigger-job
+    parent: downstream-va-hci-trigger-job
+    description: |
+      This job will trigger a VA HCI job downstream.
+      We can invoke this job by adding following comment
+      `\trigger github-experimental` in the github pull request comment.
+      These jobs runs in github-experimental-trigger pipeline.
+    files:
+      - ^lib
```
cjeanner, line 10: we should probably start having one file per DT/VA, instead of the main "defaults.yml" nested in the automation tree. That would allow to trigger the job upon automation file modification as well.

@fultonj @abays IIRC the first proposal for the automation description file was, indeed, to get one file per VA/DT. IMHO, having everything in one single file was a mistake - we can probably correct that without the need to break all of the existing jobs, by ensuring the filename matches the scenario name.

We'd then "just" need a simple change in the framework to consume the file, and voilà.

Any thoughts?

Otherwise, the file listed here looks OK, with the current automation layout.

pablintino: Some VA/DTs already have dedicated files, so I think we can move the ones to a dedicated file for each VA/DT without any issue.

abays: I have no objections to having a separate file per VA/DT.

fultonj: If we get a PR which proposes separate file per VA/DT, then I'd be happy to help merge it.

fultonj: As a way forward can do this?

- ci-framework patch and architecture patch to split default into multiple files
- then we merge this patch

raukadah: Above approach looks good to me. I will work on splitting the defaults changes and then merge this patch.

abays:
recheck-gate

PR #376, opened 2024-08-27
Add NFS network for Manila to uni04delta

fultonj:
Manila Tempest tests need to connect to the share for Ganesha via a special (openstack) network [1].

This patch adds the NFS storage network with VLAN 24 and range 172.21.0.0/24 in uni04delta. The NFS network is connected to Ceph and Compute EDPM nodes. A NNCP, NAD, L2Advertisement and IPAddressPool are defined for the NFS network so that a pod in k8s can connect to it; such as the tempest pod which will perform the storage tests and the manilaShares pod(s).

In order to make these changes, uni04delta now keeps its own copy of the nncp and networking directories since they differ (by the new network) from the generic ones in the lib directory.

[1] https://opendev.org/openstack/manila-tempest-plugin/src/branch/master/manila_tempest_tests/config.py#L99

Jira: https://issues.redhat.com/browse/OSPRH-7417
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2273

### DIFF
--- a/dt/uni04delta/kustomization.yaml
+++ b/dt/uni04delta/kustomization.yaml
@@ -17,9 +17,9 @@ transformers:
         create: true
 
 components:
-  - ../../lib/networking/metallb
-  - ../../lib/networking/netconfig
-  - ../../lib/networking/nad
+  - networking/metallb
+  - networking/netconfig
+  - networking/nad
   - ../../lib/control-plane
 
 patches:
--- a/dt/uni04delta/networking/netconfig/netconfig.yaml
+++ b/dt/uni04delta/networking/netconfig/netconfig.yaml
@@ -1,0 +1,38 @@
+---
+apiVersion: network.openstack.org/v1beta1
+kind: NetConfig
+metadata:
+  name: netconfig
+  namespace: openstack
+spec:
+  networks:
+    - dnsDomain: _replaced_
+      name: ctlplane
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: internalapi
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: external
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: storage
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: tenant
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: nfs
+      subnets:
+        - _replaced_
+      mtu: 1500
--- a/dt/uni04delta/nncp/kustomization.yaml
+++ b/dt/uni04delta/nncp/kustomization.yaml
@@ -1,0 +1,522 @@
+---
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+resources:
+  - ocp_nodes_nncp.yaml
+
+patches:
+  - target:
+      kind: NodeNetworkConfigurationPolicy
+      labelSelector: "osp/nncm-config-type=standard"
+    path: ocp_node_template.yaml
+
+replacements:
+  # Common network interfaces and vlans
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.internalapi.base_iface
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=internalapi].vlan.base-iface
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.internalapi.vlan
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=internalapi].vlan.id
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.internalapi.mtu
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=internalapi].mtu
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.tenant.base_iface
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=tenant].vlan.base-iface
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.tenant.vlan
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=tenant].vlan.id
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.tenant.mtu
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=tenant].mtu
+  - source:
+      kind: ConfigMap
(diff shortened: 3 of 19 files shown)


fultonj:
When I run the kustomize commands documented at:

https://github.com/openstack-k8s-operators/architecture/tree/main/examples/va/hci#stages

They produce the desired k8s manifests which are the same but have new storagenfs network:

https://paste.openstack.org/show/bM6mLzUtCv0OFpjgBV4r

Though, I do see VLAN 21 where 24 was desired on line 80 above so I need to fix that next.

Also, maybe I need to update the CI tests to work with the change?

fultonj:
> @fultonj in theory this additional network is required in uni-delta (external ceph + manila w/ ganesha). Should we just add it in that particular scenario or is your plan to have it to both (uni-gamma/va1 - uni-delta).

OK, thanks. I'll move it to uni-delta

fultonj:
> > @fultonj in theory this additional network is required in uni-delta (external ceph + manila w/ ganesha). Should we just add it in that particular scenario or is your plan to have it to both (uni-gamma/va1 - uni-delta).
> 
> OK, thanks. I'll move it to uni-delta

done

dt/uni04delta/networking/netconfig/netconfig.yaml
```diff
@@ -0,0 +1,33 @@
...
+      name: external
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
+      name: storage
+      subnets:
+        - _replaced_
+      mtu: 1500
+    - dnsDomain: _replaced_
```
fultonj, line 29: Should `storagenfs` be added to this list?

dt/uni04delta/nncp/kustomization.yaml
```diff
@@ -0,0 +1,523 @@
...
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.node_0.storagenfs_ip
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+          name: node-0
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=storagenfs].ipv4.address.0.ip
```
fultonj, line 197:
```
  stderr: 'error: accumulating components: accumulateDirectory:
  "recursed accumulation of path
  ''/home/zuul/src/github.com/openstack-k8s-operators/architecture/dt/uni04delta/nncp'':
  fieldPath `data.node_0.storagenfs_ip` is missing for replacement source
  ConfigMap.[noVer].[noGrp]/network-values.[noNs]"'
```

dt/uni04delta/networking/netconfig/netconfig.yaml
fultonj: yes, I have added it.

dt/uni04delta/nncp/kustomization.yaml
fultonj: The newly proposed network name does not have "storage" in it since kustomize only supports substring matching and not full regex. Thus, "storagenfs" and "storage" match when they should not and the kustomization does not work.

fultonj:
-  stderr: 'error: accumulating components: accumulateDirectory: "recursed accumulation of path
-  ''/home/zuul/src/github.com/openstack-k8s-operators/architecture/dt/uni04delta/nncp'':
-  fieldPath `data.node_0.nfs_ip` is missing for replacement source
-  ConfigMap.[noVer].[noGrp]/network-values.[noNs]"'

https://softwarefactory-project.io/zuul/t/rdoproject.org/build/93ae5b24b22944eab725b1c0db73bc01

fultonj:
recheck

fultonj:
recheck

fultonj:
Todo:

With this change `kustomize build examples/va/hci` produces the following:

```yaml
      manilaShares:
        share1:
          networkAttachments:
          - storage
```
but `nfs` should be in the list of `networkAttachments`.

fultonj:
> Todo:
> 
> With this change `kustomize build examples/dt/uni04delta` produces an `OpenStackControlPlane` with the following:
> 
> ```yaml
>       manilaShares:
>         share1:
>           networkAttachments:
>           - storage
> ```
> 
> but `nfs` should be in the list of `networkAttachments`.
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L127-L130

The following change adds `nfs` to the `networkAttachments` list.

https://github.com/openstack-k8s-operators/architecture/compare/4cb3db3c8dda30d1bba7e458cc7d465d454d5f95..c590dafbdf33e8d7e5b068e2f0839cea63a125e1

dt/uni04delta/kustomization.yaml
```diff
@@ -17,9 +17,9 @@ transformers:
         create: true
 
 components:
-  - ../../lib/networking/metallb
-  - ../../lib/networking/netconfig
-  - ../../lib/networking/nad
+  - networking/metallb
```
fmount, line 20: I'm wondering (but I don't have probably enough knowledge) if, instead of duplicating the components we used to have in `lib/` there's a way to extend them as we've done w/ `storagemgmt` for Ceph.
The nfs use case is a similar example.
The only use case where we need that network in the ctlplane is `test-operator` so far, while it's required on the edpm nodes.

fmount: Ok, while I was looking at the equivalent RGW/Swift change, I found an already started conversation around this topic  [1]. No need to duplicate answers.

[1] https://github.com/openstack-k8s-operators/architecture/pull/386#issuecomment-2340198597

dt/uni04delta/kustomization.yaml
```diff
@@ -37,6 +37,15 @@ patches:
             - _replaced_
           mtu: 1500
 
+  - target:
+      version: v1beta1
+      kind: OpenStackControlPlane
+      name: controlplane
+    patch: |-
+      - op: add
+        path: /spec/manila/template/manilaShares/share1/networkAttachments/-
```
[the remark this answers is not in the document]
fultonj: If the manila share pod does not need to be able to access the `cifmw_cephadm_nfs_vip`, then I can remove this line. Thanks.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/cephnfs.yml#L20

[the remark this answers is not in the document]
fmount: correct, we don't need to add this additional nad, so I guess L40-L48 can be removed.
`cifmw_cephadm_nfs_vip` is required on the Ceph side (hence we need it on the edpm nodes), and that network should be "abstracted" at tenant level to allow the workload mount the share using `cifmw_cephadm_nfs_vip`.
Manila itself just need to connect to ceph, and the driver actually uses the `ceph nfs ...` interface (mgr plugin) to do crud operations against ganesha (this removed the old dbus dependency we had in TripleO).

fultonj:
todo:
- streamline the kustomize per andrew's example

examples/dt/uni04delta/kustomization.yaml
```diff
@@ -9,3 +9,13 @@ resources:
...
+
+patches:
+  - target:
+      version: v1beta1
+      kind: OpenStackControlPlane
+      name: controlplane
+    patch: |-
+      - op: add
+        path: /spec/manila/template/manilaShares/share1/networkAttachments/-
+        value: nfs
```
gouthampacha, line 21: Not necessary for now; as discussed [elsewhere](https://github.com/openstack-k8s-operators/architecture/pull/376/files#r1751682609). May need to revisit this in the future if we decide to support data migrations facilitated by a control plane service.

dt/uni04delta/kustomization.yaml
fultonj: removed

examples/dt/uni04delta/kustomization.yaml
fultonj: thanks. removed.

fultonj:
Obsoleted by https://github.com/openstack-k8s-operators/architecture/pull/405

PR #379, opened 2024-08-29
Ensure we have one scenario per automation file

cjeanner:
This is in addition to a coming PR, #383. This change here is allowing
to ensure we follow those practices:
- only one scenario per file
- filename matches "scenario_name.yaml" pattern


cjeanner:
/override ci/prow/images
/override ci/prow/unit

Prow has some issues. The tests aren't relevant for this PR anyway.

PR #382, opened 2024-08-30
Introduce "adoption" snippet in automation file

cjeanner:
For adoption, there's a need to override the infra exposed by the
scenario.
While this new "adoption" entry creates a stronger link with the
ci-framework, it can also be consumed by any other tool, as long as it
knows how to consume the two "patches".

This is part of a PoC, and will probably lead to discussions regarding
the best way to manage all of the datasets.


automation/vars/hci.yaml
```diff
@@ -76,3 +76,69 @@ vas:
           - name: edpm-deployment-values-post-ceph
             src_file: values.yaml
         build_output: deployment-post-ceph.yaml
+
+# OSP related content for adoption
+osp:
```
jistr, line 81: Maybe we should call this directly "adoption" because we don't care about any "generic OSP configs" outside of adoption context.

automation/vars/hci.yaml
```diff
@@ -76,3 +76,78 @@ vas:
...
+  hci:
+    layout_patch:
+      vms:
+        # Let's remove the default computes, since we want to adopt the
+        # OSP ones
+        compute:
+          amount: 0
+        osp-undercloud:
+          amount: 1
+          disk_file_name: rhel-8.10.qcow2
```
fmount, line 91: I'm wondering if we want this level of detail in architecture, where we define a bunch of VMs settings.
I would have expected to have this kind of definition in ci-framework (like we do for greenfield, right?) and reference a "scenario" that can be loaded.
However, I'm not saying this is a bad approach, as we have a clear view of the infra associated to VA1 in the adoption context in architectures.
I'll let @marios @jistr @cescgina comment further on this.

cjeanner: we can. It could point to a thing to inherit from, but there are some caveats there, like the file location (remote vs ansible-controller). Indeed, this might be linked a bit too much to the framework, while having a description file could do.

And if we go down that path, we can push everything in that file - leading to this automation, in architecture, to have "just" `adoption: relative/path/to/scenario.yml`
But there would still be the "where to store t-h-t static/dynamic content that needs to be injected"... And there, I'm sure it doesn't belong to the ci-framework. Probably "bad" for architecture. But I'm not sure if there would be any other "good" place :/

fmount: I think that a VA/DT already defines a set of services and it can be used as source of truth to define a static set of THT env files. That part is not "strictly" connected to the infra IMO, and it might live here in terms of "description" of the source cloud.

WRT the infra part, we could minimize the changes here by referencing a scenario that lives in the `adoption` ci-framework path, but I'd like to hear more opinions.
My comment started from the assumption that for greenfield VA/DTs we have scenario files that live in ci-framework,  and we might want to be consistent with the already existing approach.

automation/vars/hci.yaml
```diff
@@ -76,3 +76,78 @@ vas:
           - name: edpm-deployment-values-post-ceph
             src_file: values.yaml
         build_output: deployment-post-ceph.yaml
+
+# Adoption related content for adoption
+adoption:
+  hci:
+    layout_patch:
```
jistr, line 83: Just diving a bit deeper into CI framework (this is my first real encounter with it).

If this is patching `cifmw_libvirt_manager_layout` then should we name the patch variables accordingly? Like `cifmw_libvirt_manager_layout_patch`? I wouldn't be afraid of long names for the sake of clarity.

Also where does the actual `cifmw_libvirt_manager_layout` value come from for the architecture jobs? I was wondering about the patching approach in general, vs. just redefining the variable for adoption, but i can't find the variable for greenfield here in this repo or in ci-framework. Is there another repo i should be looking at?

automation/vars/hci.yaml
fultonj: What if this infra style change to hci.yaml were in `ci-framework/scenarios/adoption/hci.yaml`?

Then in `architecture/automation/vars/hci.yaml` it could have a reference to `ci-framework/scenarios/adoption/hci.yaml`?

automation/vars/hci.yaml
jistr: FWIW i was looking at https://github.com/openstack-k8s-operators/ci-framework/pull/2285/files#diff-8599ae825f8eabc23397e4f5dd09e71bd037657b1f3e837565cb162d245b3a1cR93 . I guess `_cifmw_libvirt_manager_layout` there gets processed from `cifmw_libvirt_manager_configuration`? https://github.com/openstack-k8s-operators/ci-framework/blob/b349f2735319daea1d1b43cb6b14cd7af4fa16db/scenarios/reproducers/va-hci.yml#L21

cjeanner: The patching is done here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/libvirt_manager/tasks/generate_layout.yml

patches are nice, since we don't need to redefine the whole thing. There's one weakness though : we can't delete any entry.

The playbooks I proposed in https://github.com/openstack-k8s-operators/ci-framework/pull/2285 are taking care of the dataset manipulation/patching

automation/vars/hci.yaml
cjeanner: we could, indeed. We'd even NOT need any reference from any file: the scenario name, and the fact it's "adoption" should be sufficient to built the pointer to the file.

It won't solve the "where do we put OSP t-h-t content", but it reduces the amount of affected repositories in case of changes. I didn't think about it, and I'm ashamed of that ;_;.

Far, far cleaner to pass via scenario files - and that's the purpose if them anyway!

cjeanner:
Closing this PR, since we can now rely on the ci-framework scenario nested in `scenarios/adoption/` tree, at least for the adoption related overrides.

You can check https://github.com/openstack-k8s-operators/ci-framework/pull/2285 to follow the evolution.

PR #388, opened 2024-09-05
Add 'local' network type in the Control Plane CRs

slawqo:
Lack of this 'local' network type breaks somehow some Horizon integration tests as this is first choice in the Horizon UI. From Neutron PoV there is no reason not to load this driver so lets enable that driver in the Neutron's config.

### DIFF
--- a/examples/dt/uni01alpha/control-plane/service-values.yaml
+++ b/examples/dt/uni01alpha/control-plane/service-values.yaml
@@ -173,5 +173,5 @@ data:
       ovn_emit_need_to_frag = true
 
       [ml2]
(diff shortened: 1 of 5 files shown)


fultonj:
@slawqo since downstream ci uses 18.0.0-proposed, if you want this tested downstream please cherry pick as per:

https://github.com/openstack-k8s-operators/architecture/blob/main/docs/contributing/cherry-picking.md

slawqo:
thx @fultonj I just proposed backport https://github.com/openstack-k8s-operators/architecture/pull/389

PR #389, opened 2024-09-05
Add 'local' network type in the Control Plane CRs

slawqo:
Lack of this 'local' network type breaks somehow some Horizon integration tests as this is first choice in the Horizon UI. From Neutron PoV there is no reason not to load this driver so lets enable that driver in the Neutron's config.

(cherry picked from commit 82b8592168adfb64784544991b22601502289343)


fultonj (approved):
/approve
/lgtm

fultonj (approved):
/approve
/lgtm

PR #391, opened 2024-09-06
[delta-ipv6] Shared control plane vars

afazekas:
There is 2 stage control plane config,
the shared options moved to different file used by both stages.


hjensas (commented):
/lgtm

abays (approved):
/lgtm
/approve

PR #400, opened 2024-09-17
[DNM] Force Ansible debug verbosity for VA1

abays:
Needed for uni03gamma debugging


abays:
This shouldn't be needed anymore

PR #378, opened 2024-08-29
Multiple nodesets for ovs-dpdk-sriov scenario

mnietoji:
Depends-on: https://github.com/openstack-k8s-operators/ci-framework/pull/2541
Created a new dt based on ovs-dpdk-sriov va

PR #401, opened 2024-09-17
Add multi-cell DT

jamepark4:
Validating a multi-cell deployment.


examples/dt/nova/nova-multiplecells/values.yaml
```diff
@@ -0,0 +1,130 @@
...
+apiVersion: v1
+kind: ConfigMap
+
+metadata:
+  name: edpm-nodeset-values
+  annotations:
+    config.kubernetes.io/local-config: "true"
+
+data:
+  ssh_keys:
```
hjensas, line 11: I am not a fan of hacing ssh_keys set in two values files in this PR. The nodeset in the library will create the same secret `dataplane-ansible-ssh-private-key-secret` for both nodesets. If a user looks at this they may be led to believe they can use different keys for the (I assume pre-provisioned nodes?) when using this DT - but that will not work since applying the second nodeset will modify the same secret instead of creating a new one.

jamepark4: I'm not familiar with how the common template interacts with dt/va variant templates.  If the ssh keys are not present in the DT/VA's will [common](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2#L14) add in the keys?

jamepark4: Ack I can removed these.

hjensas: I am not sure you can remove them, if you cannot - then this change should not be blocked by this comment from me.

I was trying to point out an issue with the library not supporting multiple nodesets in a good way. Essentilly you have two nodesets referencing lib/dataplane/nodeset/kustomization.yaml#L6-L8 - so you end up applying two separate files defining the same secret resource.

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
```diff
@@ -0,0 +1,83 @@
...
+          conductorServiceTemplate:
+            replicas: 3
+          hasAPIAccess: true
+        cell2:
+          cellDatabaseAccount: nova-cell2
+          cellDatabaseInstance: openstack-cell2
+          cellMessageBusInstance: rabbitmq-cell2
+          conductorServiceTemplate:
+            replicas: 3
+          hasAPIAccess: true
```
gibizer, line 83: Just a note for the future. We can spice this setup up by defining cell2 without api DB access (i.e. no upcall support).

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
```diff
@@ -0,0 +1,83 @@
...
+  telemetry:
+    enabled: true
+    template:
+      ceilometer:
+        enabled: true
+
+  rabbitmq:
+    templates:
+      rabbitmq-cell2:
+        replicas: 3
```
gibizer, line 57: Is this intentional that for galera above all the 3 DBs are re-defined, but for rabbitmq here only the additional MQ is defined?

examples/dt/nova/nova-three-cells/edpm/nodeset2/values.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - install-os
+      - configure-os
+      - ssh-known-hosts
+      - run-os
+      - reboot-os
+      - install-certs
+      - ovn
+      - neutron-metadata
+      - libvirt
+      - nova
```
gibizer, line 124: wondering how this connects the nodes to cell2 if this still refers to the default nova service.

examples/dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,185 @@
...
+cells
+
+## Environment
+
+### Nodes
+
+| Role              | Machine Type | Count |
+| ----------------- | ------------ | ----- |
+| Compact OpenShift | vm           |   3   |
+| OpenStack Compute | vm           |   4   |
```
gibizer, line 18: Does this DT really deploys 4 EDPM nodes as computes? I only see edpm-compute-0 and edpm-compute-1 referenced in the NodeSets.

examples/dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,185 @@
+# Deployed Topology - Nova/MultipleCells
+
+Deploys an environment with three cells instead of the standard two cells
+
+## Purpose
+
+Focused on leveraging multiple nodesets to deploy two distinct cells 
+(cell1 and cell2) in addition to cell0. Computes will be allocated to both 
+cells
```
gibizer, line 9: Would be nice to describe the mapping. Which compute will be connected to which cell.

gibizer (changes requested):
I have couple of question / comments inline.

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
bogdando: Please use a separate MQ cluster for each cell

examples/dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+Increase default cells to include cell2
+
+```YAML
+---
+spec:
+  nova:
+    template:
+      secret: osp-secret
+      cellTemplates:
```
bogdando, line 83: please define the metadataServiceTemplate, either to run it globally, or per a cell
See https://github.com/bogdando/data-plane-adoption/blob/f61537f9701f6585183810b20e2142423ae4f8d0/docs_user/modules/proc_adopting-the-compute-service.adoc#L51 and https://github.com/bogdando/data-plane-adoption/blob/f61537f9701f6585183810b20e2142423ae4f8d0/docs_user/modules/proc_adopting-the-compute-service.adoc#L95 examples

If running it per a cell, we will also need a generated secret for neutron https://github.com/bogdando/data-plane-adoption/blob/f61537f9701f6585183810b20e2142423ae4f8d0/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc#L285

examples/dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,185 @@
...
+metadata:
+  name: nova-cell-2
+  namespace: openstack
+spec:
+  addCertMounts: false
+  caCerts: combined-ca-bundle
+  containerImageFields:
+  - NovaComputeImage
+  - EdpmIscsidImage
+  dataSources:
```
bogdando, line 174: here we should add that ``nova-cell<N>-metadata-neutron-config``, if running metadata service in a cell instead of top scope

examples/dt/nova/nova-three-cells/edpm/nodeset2/values.yaml
bogdando: right, that won't work, please see my WIP for adoption, perhaps you may find somthing to reuse https://github.com/bogdando/data-plane-adoption/blob/f61537f9701f6585183810b20e2142423ae4f8d0/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc#L237 , then how we use it in the services list https://github.com/bogdando/data-plane-adoption/blob/f61537f9701f6585183810b20e2142423ae4f8d0/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc#L381

examples/dt/nova/nova-three-cells/values.yaml
```diff
@@ -0,0 +1,130 @@
...
+      - install-os
+      - configure-os
+      - ssh-known-hosts
+      - run-os
+      - reboot-os
+      - install-certs
+      - ovn
+      - neutron-metadata
+      - libvirt
+      - nova
```
bogdando, line 130: ditto, use ``nova-<cell>`` in ``<cell>`` specific nodesets

examples/dt/nova/nova-three-cells/edpm/nodeset2/values.yaml
bogdando: oh indeed. replacing via kustomize works for me

examples/dt/nova/nova-three-cells/values.yaml
bogdando: perhaps this is also replaced via kustomize?

examples/dt/nova/nova-three-cells/README.md
jamepark4: Because this is completely virtualized its unfortunately not very clear based on purely this DT.  This DT works in conjunction with templates in CIFMW [1] that splits the total computes between each nodeset.   It also needs a scenario file similar to [2] to define the number of VMs needed (the one I'm using for testing this is define downstream).  Virtualized setups now have unique hostnames that are recorded during setup and that information is what [1] uses to allocate how the hosts are split.  With baremetal setups where I can define the hostnames in the jobs, I can specifically define how the EDPM Hosts are allocated per nodeset within the context of the DT.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2423/files
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/va-nfv-ovs-sriov.yml

examples/dt/nova/nova-three-cells/README.md
jamepark4: Ack

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
jamepark4: @gibizer I need to take a look at this again, you actually helped me debug this back in early October.  The deployment was requesting something like 500M by default so I updated all of them to 5G.  Although looking at the architecture library it should be using 5G.  Let me revisist why this was happening and I'll reduce if possible.

@bogdando ack, thanks.

examples/dt/nova/nova-three-cells/README.md
jamepark4: ack thanks, let me try and get a per cell working.

examples/dt/nova/nova-three-cells/README.md
jamepark4: ack I'll try for per cell deployment.

examples/dt/nova/nova-three-cells/values.yaml
jamepark4: I was only updating cell2 and keeping the default for cell1 so this isn't replaced, but if it improves overall clarity I can update this and kustomize to have `nova-cell-1` and `nova-cell-2` respectively instead.

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
jamepark4: @bogdando When you say separate MQ for each cluster do you mean unique metallb annotations per rabbitmq cluster?  If so I'm currently getting those via CIFMW.  I can add that as part of the README to help clarify.

```
  rabbitmq:
    enabled: true
    templates:
      rabbitmq:
        delayStartSeconds: 30
        override:
          service:
            metadata:
              annotations:
                metallb.universe.tf/address-pool: internalapi
                metallb.universe.tf/loadBalancerIPs: 172.17.0.85
...
      rabbitmq-cell1:
        delayStartSeconds: 30
        override:
          service:
            metadata:
              annotations:
                metallb.universe.tf/address-pool: internalapi
                metallb.universe.tf/loadBalancerIPs: 172.17.0.86
...
      rabbitmq-cell2:
        delayStartSeconds: 30
        override:
          service:
            metadata:
              annotations:
                metallb.universe.tf/address-pool: internalapi
                metallb.universe.tf/loadBalancerIPs: 172.17.0.87
```

gibizer: I think the current proposal already uses individual rabbit for each cell. 
cell0: rabbit-server
cell1: rabbit-cell1-server
cell2: rabbit-cell2-server
So I think we are good.

examples/dt/nova/nova-three-cells/edpm/nodeset2/values.yaml
gibizer: sorry, yeah kustomize is OK to me too.

examples/dt/nova/nova-three-cells/README.md
bogdando: please fill a bug - that is users facing manifest, it must contain explicit things, including metadataServiceTemplate (or providing it in kustomizations). However, ci-framework is NOT users facing, and doing implicit things there renders these manifests unusable outside of CI infra

gibizer: The question for me is the following: Does running kustomize on this DT locally will produce a complete CR? If so then I'm happy. If it does not produce complete CR but left out metadataTemplate definition then it is a bug indeed.

jamepark4:
recheck

jamepark4:
> @jamepark4 Would you mind addressing the errors reported in CI when you have a chance?
Sure I need to make some changes to [1] and then get it merged in order for the validations evaluating the kustomize for nodeset2 to work.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2423

jamepark4:
@abays CI errors have been addressed, do you mind looking this over and providing any input/reviews if you have the bandwidth?

dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,11 @@
+# Deployed Topology - Nova/MultipleCells
+
+If you are looking for information on how to deploy the multiplecells based DT, then
+please the [README](../../examples/dt/nova/nova- multiplecells/README.md) in the examples
+directory.
+
+This directory `dt/nova/multiplecells/`, exists so that the
+[kustomization.yaml](../../examples/dt/nova/nova-three-cells/kustomization.yaml) in
```
abays, line 8: This link gives a 404

dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,11 @@
+# Deployed Topology - Nova/MultipleCells
+
+If you are looking for information on how to deploy the multiplecells based DT, then
+please the [README](../../examples/dt/nova/nova- multiplecells/README.md) in the examples
```
abays, line 4: This `[README]...` renders as literal text instead of creating a hyperlink

examples/dt/nova/nova-three-cells/data-plane.md
```diff
@@ -0,0 +1,2 @@
+# Deploying the OpenStack dataplane
+TBD
```
abays, line 2: Did you forget to finish this section?

dt/nova/nova-three-cells/README.md
jamepark4: Ack

dt/nova/nova-three-cells/README.md
jamepark4: Ack

examples/dt/nova/nova-three-cells/data-plane.md
jamepark4: This and controlplane, thanks for catching this.

dt/nova/nova-three-cells/edpm/nodeset/nova_cell_service.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: nova
```
gibizer, line 5: Don't we have this automatically created by the openstack-operator based on the built in default service definitions in https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml ?

Does other DT/VAs define the nova service the same way so you are just following a pattern or is it something specific to this DT?

Is it only needed here to be able to create a copy of it and patch is to define the cell2 nova service? If so then I'm wondering why we don't simply store the cell2 nova service instead of patching of a copy. I could imagine that we want to keep cell1 and cell2 nova service consistent but I think we can only do so if we somehow able to reuse the default definition from https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_nova.yaml as that is the source of truth afaik.

examples/dt/nova/nova-three-cells/edpm/nodeset2/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - target:
+      kind: OpenStackDataPlaneNodeSet
+      name: openstack-edpm
+    patch: |-
+      - op: replace
+        path: /metadata/name
+        value: openstack-edpm-2
+      - op: replace
+        path: /spec/services/13
+        value: nova-cell-2
```
gibizer, line 21: This patching feels strange to me. We take default NodeSet and duplicate it for cell2 and patch it to connect the computes in it to cell2. But we 
i) don't define the computes of NodeSet/openstack-edpm-2 here, 
ii) we have a complete NodeSet  parameter definition with the specific compute in nodeset2/values.yaml 

So basically we have two ways to pass values to NodeSet/openstack-edpm2:
* here with the kustomize patch
* via nodeset2/values.yaml that is applied by another kustomize patch I assume.

This make it hard to follow what parameter comes from where. I don't want to block on this but I feel this needs to be refactored if possible to provide parameters to the NodeSet from a single source to help maintaining the kustomization. Is it something that can be done in a follow up? 

@abays how do you feel about this issue? Do you agree that this is a problem? If so do you think it can be refactored to a single source? Is it something new in the proposed DT or is it widespread across the DTs?

gibizer (commented):
I still have couple of questions / comments.

examples/dt/nova/nova-three-cells/control-plane/service-values.yaml
jamepark4: Ack I'll remove the 3 galera and keep it inline with how rabbitmq is being defined here.

examples/dt/nova/nova-three-cells/README.md
jamepark4: Yes that makes sense and you are correct the FW is overriding it to create the 4. Since that can't be clearly conveyed within just to context of the DT, I will update to 2.

examples/dt/nova/nova-three-cells/README.md
```diff
@@ -0,0 +1,183 @@
+# Deployed Topology - Nova/Three Cells
+
+Deploys an environment with three cells instead of the standard two cells
+
+## Purpose
+
+Focused on leveraging multiple nodesets to deploy two distinct cells 
+(cell1 and cell2) in addition to cell0. Computes-0/1 are allocated
+to cell1 and computes-2/3 will be placed in cell2.
```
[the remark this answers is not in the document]
jamepark4: Yes correct I'll remap based on two.

dt/nova/nova-three-cells/edpm/nodeset/nova_cell_service.yaml
jamepark4: Its the latter, you are correct the service is defined without needing to be explicitly defined in the DT.  But the only mechanism that would allow me to create a second service for cell2 would be to define it within the DT.  Let me try moving this to purely cell2 though instead of as a copy and patch.

gibizer (approved):
Thanks for the cleanups. Looks good to me now.
Also I see that computes are properly mapped to different cells in https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/6f99555d33dc4e4c9404e01703ab3aa4/log/logs/controller-0/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/ctlplane/nova/host_list

So this is good to go.

bogdando:
lgtm

fultonj:
I have no objections to this PR if @abays wants to merge it.

examples/dt/nova/nova-three-cells/edpm/nodeset2/kustomization.yaml
abays: > I feel this needs to be refactored if possible to provide parameters to the NodeSet from a single source to help maintaining the kustomization.

So what I think might happen when using this in CIFMW is the same problem that was encountered in this PR [1].  Since ultimately the `lib` dataplane nodeset is used [2] by both nodesets, they will collide with each other when CIFMW's `kustomize_deploy` role generates all of the YAMLs ahead time.  But maybe you somehow found a way around the problem.  I'll be curious to see the test results for this.  Since this doesn't change any existing, underlying `lib`, `va`, or `dt` content, I'm good with merging it for now.

[1] https://github.com/openstack-k8s-operators/architecture/pull/378
[2] https://github.com/jamepark4/architecture/blob/191d2bfabdc1355f5ddce7bf47a264fa46c12b45/dt/nova/nova-three-cells/edpm/nodeset/kustomization.yaml#L20

jamepark4: Using CIFMW downstream I'm not seeing any collisions since the automation/vars uses two different values: `edpm-nodeset-values` and `edpm-nodeset2-values`.  With nodeset2's name being updated by [1].  `kustomize_deploy` ends up generating two yaml definitions, nodeset.yaml and nodeset2.yaml.  The different Nodeset names allows them to be deployed without any issues.

[1] https://github.com/openstack-k8s-operators/architecture/pull/401/files#diff-20af9459981c1a0588160942c3490649f5cf4fd222526cea08c3244208fb48ecR13

abays: That's great news.  @mnietoji maybe you can review how two nodesets were used here and apply the same design to fix your PR (if you haven't already figured it out on your own)?

abays (approved):
/lgtm
/approve

PR #402, opened 2024-09-18
[delta-ipv6] Shared control plane vars

afazekas:
There is two stage with control plane config,
the shared options moved to a shared file which
used by both stages.


fultonj:
@afazekas is this being done to address a bug? What about this is specific to IPv6?

afazekas:
delta-ipv6 has ironic enabled and the 2th definition of the control-plane otherwise undo the ironic config.

fultonj:
+2 from me

@hjensas are you ok with this?

hjensas:
> +2 from me
> 
> @hjensas are you ok with this?

+2 from me as well, without this we ended up enabling ironic when initially deploying the controlplane, then disabling it post ceph deployment when the controlplane was re-configured.

fultonj (approved):
/approve
/lgtm

fultonj:
please consider cherry picking as per https://github.com/openstack-k8s-operators/architecture/blob/main/docs/contributing/cherry-picking.md

PR #411, opened 2024-10-01
Enable Heat for uni04delta-ipv6 DT

gmarcian:
Since Tobiko tf uses Heat, we need to enable it in order to run Tobiko tests.

### DIFF
--- a/dt/uni04delta-ipv6/control-plane/kustomization.yaml
+++ b/dt/uni04delta-ipv6/control-plane/kustomization.yaml
@@ -143,3 +143,15 @@ replacements:
           - spec.nova.template.cellTemplates
         options:
           create: true
+
+  - source:
+      kind: ConfigMap
+      name: service-values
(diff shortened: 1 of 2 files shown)


fultonj (approved):
/aprove
/lgtm

I verified that I could build delta-ipv6 with this change and diff'd the output before and after. the only diff is that heat is enabled as it should be.

fultonj:
please cherry pick  https://github.com/openstack-k8s-operators/architecture/blob/main/docs/contributing/cherry-picking.md

gmarcian:
@fultonj please see: https://github.com/openstack-k8s-operators/architecture/pull/415
BTW, may i ask to add me as a member to the /openstack-k8s-operators group?

abays:
> @fultonj please see: #415 BTW, may i ask to add me as a member to the openstack-k8s-operators org?

@gmarcian Invite sent

abays:
recheck-gate

PR #419, opened 2024-10-16
Set enable-chassis-as-gateway controlplane as false in alpha and eta

fyanac:
Alpha and eta07 have networkers that are using them as a gateway.


eduolivares:
Something I have been talking with @fyanac:
I know octavia folks are working to enable octavia and add octavia tests to some downstream uni jobs.
In order to work properly, octavia amphora provider needs OCP nodes acting as gateway chassis.
So, with this PR, octavia amphora provider could not be used in both alpha and eta jobs.

Hence, please check with octavia folks what we should do before merging this patch (maybe we can apply it only to one of the jobs, IDK...).

Ccing @lavraham @beagles @gthiemonge

fyanac:
> Something I have been talking with @fyanac: I know octavia folks are working to enable octavia and add octavia tests to some downstream uni jobs. In order to work properly, octavia amphora provider needs OCP nodes acting as gateway chassis. So, with this PR, octavia amphora provider could not be used in both alpha and eta jobs.
> 
> Hence, please check with octavia folks what we should do before merging this patch (maybe we can apply it only to one of the jobs, IDK...).
> 
> Ccing @lavraham @beagles @gthiemonge

I modified the patch just to change it into uni07eta job.

eduolivares:
you need to add a section here:
https://github.com/fyanac/architecture/blob/change_gw/dt/uni07eta/kustomization.yaml#L322-L332
for `data.ovn.ovnController.external-ids` so that kustomize applied your new values to the output CR.

It will be similar to this:
https://github.com/fyanac/architecture/blob/change_gw/examples/dt/bgp/bgp_dt01/control-plane/kustomization.yaml#L147-L158

PR #420, opened 2024-10-17
[POC] Add an adoption VA to setup networking

cescgina:
Create an adoption VA the will be used to install the operators, and
setup nncp and netconfig. The idea is that this VA will be shared
between all the different adoption uni-jobs.


abays:
This seems fine to me.  My understanding is that it's just reusing part of the existing HCI VA via a new `automation` flow.  I'm not sure I understand the job failure though.

abays (approved):
/lgtm
/approve

odyssey4me:
@fultonj Could you please review https://github.com/openstack-k8s-operators/architecture/pull/420 - we need this in place to unblock further work on adoption to the va-hci environment.

fultonj (approved):
/approve
/lgtm

PR #425, opened 2024-10-25
[BGP] Add AZs to schedule Octavia mgmt network on OCP nodes

eduolivares:
Octavia needs to schedule its management network on the OCP nodes/workers.
With BGP, OCP nodes/workers should not be used to schedule OVN router ports.
In order to resolve this conflict, this patch uses Availability Zones.


examples/dt/uni01alpha/control-plane/service-values.yaml
```diff
@@ -94,8 +94,11 @@ data:
     enabled: true
     amphoraImageContainerImage: quay.io/gthiemonge/octavia-amphora-image
     apacheContainerImage: registry.redhat.io/ubi9/httpd-24:latest
-    availabilityZones:
-      - zone-1
+    lbMgmtNetwork:
+      availabilityZones:
```
gthiemonge, line 98: ok, you're handling it here, thanks

eduolivares:
> /lgtm, is this tested already?

@karelyatin, it's not ready yet. Notice the do-not-merge label.
I will ping you when it is ready. Thanks!

PR #431, opened 2024-10-31
Add telemetry metricStorage for NFV

rsafrono:
metricStorage is required for enabling power monitoring feature that is planned in the next feature release of RHOSO. The feature is intended for NFV workloads power monitoring.


fultonj (approved):
/approve
/lgtm

PR #432, opened 2024-10-31
Update Glance edge pod replica count from 3 to 1

fultonj:
After discussion with the Glance team we agreed that pods of type edge should default to 1, not 3, replicas.

Also set the cinderVolumes glance_api_servers for az0 to https://glance-az0-internal.openstack.svc:9292.

Also add missing `backup_ceph_conf` for Cinder backup.


abays (approved):
/lgtm
/approve

PR #436, opened 2024-11-06
DCN Workaround to set Cinder uniquePodNames false

fultonj:
As per the [OSPRH-11240](https://issues.redhat.com//browse/OSPRH-11240), set uniquePodNames false for Cinder so that extraMounts can be used to propagate secrets to pods by matching the instance AZ name in DCN deployments. This patch can be reverted after [OSPRH-11240](https://issues.redhat.com//browse/OSPRH-11240) is resolved.

Jira: https://issues.redhat.com/browse/OSPRH-11240

### DIFF
--- a/examples/dt/dcn/service-values.yaml
+++ b/examples/dt/dcn/service-values.yaml
@@ -8,6 +8,8 @@ metadata:
     config.kubernetes.io/local-config: "true"
 data:
   preserveJobs: false
+  cinder:
+    uniquePodNames: false
   cinderAPI:
     replicas: 3
     customServiceConfig: |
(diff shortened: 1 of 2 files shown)


examples/dt/dcn/service-values.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
     config.kubernetes.io/local-config: "true"
 data:
   preserveJobs: false
+  cinder:
+    uniquePodNames: false
```
fmount, line 12: Ok, looks good to me. I see we are actually enabling this parameter by default in [1], and not just for NFS (unibeta).
I'm ok to minimize the amount of changes, and good to have `uniquePodNames` enabled by default, so we can catch issues earlier.
+1 for me.
 
[1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L61

fmount (approved):
/lgtm

abays (approved):
/lgtm
/approve

PR #439, opened 2024-11-08
DCN DT: Add multicells

krcmarik:
One dedicated cell per site is being deployed with dedicated rabbitmw, Galera and nova services. The nova service and neutron-metadata service are pointed to their dedicated cell per DCN site.

### DIFF
--- a/examples/dt/dcn/values.yaml
+++ b/examples/dt/dcn/values.yaml
@@ -1,8 +1,6 @@
 ---
 apiVersion: v1
 data:
-    customDataplanService:
-        name: nova-custom-ceph
     nodeset_name: computes-edpm
     ceph_conf:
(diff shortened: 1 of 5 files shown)


examples/dt/dcn/values.yaml
```diff
@@ -29,16 +27,26 @@ data:
...
     nova:
+        customDataplaneService:
+            name: nova-custom-ceph
         ceph:
             conf: CHANGEME_NOVA_CEPH_CONF
         name: ceph-nova
         dataSources:
             - configMapRef:
-              name: ceph-nova
+                  name: ceph-nova
```
fultonj, line 37: Can we not change the number of spaces here?

fultonj (approved):
/approve
/lgtm

PR #440, opened 2024-11-11
[DCN] Explicitly set network availability zone parameters

sbekkerm:
This patch explicitly sets the network availability zone parameters to their default values, enabling the configuration of availability zones for DCN sites if needed.


fultonj (approved):
/approve
/lgtm

PR #444, opened 2024-11-13
Switch to 1500 MTU consistently

sdatko:
We were having mixed values in some scenarios,
in some places specifying 9000 while in fact
1500 value was used in the deployed infrastructure.
This commit changes it consistently for related uni jobs
to stick with default 1500 MTU.

This change intentionally ommits uni04delta and uni03gamma
(va/hci scenario), as those involve Ceph, which probably
should still rely on MTU 9000, however it is not properly
configured so far as I noticed.

### DIFF
--- a/examples/dt/uni01alpha/control-plane/nncp/values.yaml
+++ b/examples/dt/uni01alpha/control-plane/nncp/values.yaml
@@ -44,7 +44,7 @@ data:
         name: subnet1
     prefix-length: 24
     iface: enp7s0
-    mtu: 9000
+    mtu: 1500
     lb_addresses:
       - 192.168.122.80-192.168.122.90
     endpoint_annotations:
@@ -108,7 +108,7 @@ data:
         cidr: 172.18.0.0/24
         name: subnet1
         vlan: 21
-    mtu: 9000
+    mtu: 1500
     prefix-length: 24
     iface: storage
     vlan: 21
--- a/examples/dt/uni02beta/control-plane/nncp/values.yaml
+++ b/examples/dt/uni02beta/control-plane/nncp/values.yaml
@@ -47,7 +47,7 @@ data:
         name: subnet1
     prefix-length: 24
     iface: enp6s0
-    mtu: 9000
+    mtu: 1500
     lb_addresses:
       - 192.168.122.80-192.168.122.90
     endpoint_annotations:
@@ -109,7 +109,7 @@ data:
         cidr: 172.18.0.0/24
         name: subnet1
         vlan: 21
-    mtu: 9000
+    mtu: 1500
     prefix-length: 24
     iface: storage
     vlan: 21
@@ -138,7 +138,7 @@ data:
         cidr: 172.20.0.0/24
         name: subnet1
         vlan: 23
-    mtu: 9000
+    mtu: 1500
 
   tenant:
     dnsDomain: tenant.example.com
--- a/examples/dt/uni06zeta/control-plane/nncp/values.yaml
+++ b/examples/dt/uni06zeta/control-plane/nncp/values.yaml
@@ -41,7 +41,7 @@ data:
         name: subnet1
     prefix-length: 24
     iface: enp6s0
-    mtu: 9000
+    mtu: 1500
     lb_addresses:
       - 192.168.122.80-192.168.122.90
     endpoint_annotations:
@@ -103,7 +103,7 @@ data:
         cidr: 172.18.0.0/24
         name: subnet1
         vlan: 21
-    mtu: 9000
+    mtu: 1500
     prefix-length: 24
     iface: storage
     vlan: 21
--- a/examples/dt/uni07eta/control-plane/nncp/values.yaml
+++ b/examples/dt/uni07eta/control-plane/nncp/values.yaml
@@ -42,7 +42,7 @@ data:
         name: subnet1
     prefix-length: 24
     iface: enp6s0
-    mtu: 9000
+    mtu: 1500
     lb_addresses:
       - 192.168.122.80-192.168.122.90
     endpoint_annotations:
@@ -106,7 +106,7 @@ data:
         cidr: 172.18.0.0/24
         name: subnet1
         vlan: 21
-    mtu: 9000
+    mtu: 1500
     prefix-length: 24
     iface: storage
     vlan: 21


fultonj:
I don't like this change because...

The storage and storage_mgmt network benefit from having MTU 9000 with Ceph. You would not deploy Ceph without setting the MTU to 9000. Any VA/DT which uses Ceph should have this or it's less realistic. 

Because in our deployment the Storage and Storage Mgmt network are VLANs trunked from control plane, control plane also needs to be MTU 9000 if we want Storage and Storage Mgmt to use MTU 9000.

sdatko:
> The storage and storage_mgmt network benefit from having MTU 9000 with Ceph. You would not deploy Ceph without setting the MTU to 9000. Any VA/DT which uses Ceph should have this or it's less realistic.

So, uni01alpha, uni02beta, uni06zeta and uni07eta do not involve Ceph.

> Because in our deployment the Storage and Storage Mgmt network are VLANs trunked from control plane, control plane also needs to be MTU 9000 if we want Storage and Storage Mgmt to use MTU 9000.

So, in uni05epsilon we did not set MTU 9000 before and in uni04delta it was specified here, but not actually used I think (at least according to my quick check, but I may be wrong).

fultonj:
> > The storage and storage_mgmt network benefit from having MTU 9000 with Ceph. You would not deploy Ceph without setting the MTU to 9000. Any VA/DT which uses Ceph should have this or it's less realistic.
> 
> So, uni01alpha, uni02beta, uni06zeta and uni07eta do not involve Ceph.

I agree and I have no problem with them having the 1500 MTU.

> > Because in our deployment the Storage and Storage Mgmt network are VLANs trunked from control plane, control plane also needs to be MTU 9000 if we want Storage and Storage Mgmt to use MTU 9000.
> 
> In uni05epsilon we did not set MTU 9000 before and in uni04delta it was specified here, but not actually used I think (at least according to my quick check, I did not see 9000 in any network interface, but I may be wrong); the same was for uni03gamma (VA/HCI).

Do you want to remove your changes to the following from this patch?

examples/va/hci/edpm-pre-ceph/nodeset/values.yaml
examples/va/hci/control-plane/nncp/values.yaml

so that your change only affects items in the `dt` directory?

sdatko:
@abays @fultonj ack, altered the merge request to only affect uni-jobs (without va/hci), as requested.

Although, I still believe the current values for va/hci (AKA uni03gamma) are ineffective, as the scenario file https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/va-hci.yml do not specify the MTU for libvirt, so the existing entries given in archiecture repository are misleading...

examples/dt/uni04delta-ipv6/control-plane/nncp/values.yaml
abays, line 1: Shouldn't `uni04delta-ipv6` keep the `9000` MTU since it's using Ceph [1] on the storage network?  Same goes for `uni04detla` (non-ipv6) [2].

[1] https://github.com/openstack-k8s-operators/architecture/blob/1100a5097c0eab410228248b192419935d573113/automation/vars/uni04delta-ipv6.yaml#L51-L55
[2] https://github.com/openstack-k8s-operators/architecture/blob/1100a5097c0eab410228248b192419935d573113/automation/vars/uni04delta.yaml#L51-L55

sdatko: Done – I removed change for uni04delta and will create a dedicated task to investigate the missing configuration on the libvirt/infra side.

sdatko:
Meh, I was just trying my luck :laughing:

fultonj:
> @abays @fultonj ack, altered the merge request to only affect uni-jobs (without va/hci), as requested.
> 
> Although, I still believe the current values for va/hci (AKA uni03gamma) are ineffective, as the scenario file https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/va-hci.yml do not specify the MTU for libvirt, so the existing entries given in archiecture repository are misleading...

I think you think it's misleading because you're only looking at architecture from a CI point of view.

If someone follows our written directions here:

  https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/control-plane.md#initialize

and then goes to modify the values.yaml file for their environment, which doesn't use ci-framework but deploys on bare metal, that they would do well to leave the MTU set as in the example at 9000:

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/control-plane/nncp/values.yaml#L44

I realize that the uni job gamma tests va/hci. If you wish to run gamma, the DT, with MTU 1500, then its MTU change should be made by overriding a ci-framework job. The reproducers/va-hci.yml file is part of this.

fultonj:
I'm +2 on this PR as is (obviously it needs to pass all the tests before merging). My most recent comment was just to explain why I didn't want the examples yaml in the VA altered by this PR.

fultonj:
Shouldn't this patch also update the MTU for delta?

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni04delta/values.yaml#L87
https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni04delta/control-plane/nncp/values.yaml#L45

It's important to understand that processing occurs to substitute values from the ci-framework environment here:

https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_gen_kustomize_values

fultonj:
> Shouldn't this patch also update the MTU for delta?
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni04delta/values.yaml#L87 https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni04delta/control-plane/nncp/values.yaml#L45

Per our discussion, we don't have to patch delta and can merge this as is if needed.

> It's important to understand that processing occurs to substitute values from the ci-framework environment here:
> 
> https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/ci_gen_kustomize_values

Per our discussion, since the above is what swaps in new values from the ci-framework environment we may not really need this PR as the MTU values are replaced by the above role.

fultonj:
- I'm fine with merging this PR since the jobs which are changed do not need MTU 9000 and can use 1500
- This PR shouldn't really affect the jobs however since they get these values from ci-framework

sdatko:
Let's merge this as it is and make it closer to what we daily test with uni01alpha, uni02beta, uni06zeta and uni07eta.

abays (approved):
/lgtm
/approve

PR #448, opened 2024-11-18
[OSPRH-12761] Support new OpenStack initialization resource for RHOSO install

abays:
We have a new upcoming way to install and deploy the OSP operators.  OLM will only install one CRD for OSP, which is an initialization resource.  A CR must be created for that CRD, which then, when created, causes all the remaining OSP CRDs to be installed and also deploys their respective operators.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1185

### DIFF
--- a/examples/common/README.md
+++ b/examples/common/README.md
@@ -3,10 +3,9 @@
 All VAs and DTs assume that these CRs have been created.
 
 ## OLM
-
 The [olm](olm) directory contains a kustomization which will generate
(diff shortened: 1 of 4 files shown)


pinikomarov:
looks ok to me

abays:
Note: do NOT backport this to FR1.

fultonj (approved):
/lgtm
/approve

fultonj (approved):
/approve
/lgtm

PR #459, opened 2024-12-03
Set Cinder customServiceConfig globally in DCN DT

fultonj:
The customServiceConfig for cinder in the DCN DT is used to set the storage_availability_zone. We wish to do this for all Cinder services in the default site (`az0`), not just for cinderAPI or cinderBackup as the AZ was missing from the cinder-scheduler. Thus, we will move it to the global section under the cinder template and out of the cinderAPI and cinderBackup since they will inherit it. We still need to set the backend_availability_zone in all cinderVolumes except az0 since it will also inherit the new default value. However, the additional line is kept in az0's customServiceConfig to make the pattern easier to see for az0, az1, az2, etc.

Jira: https://issues.redhat.com/browse/OSPRH-11915

### DIFF
--- a/dt/dcn/edpm-post-ceph/nodeset/kustomization.yaml
+++ b/dt/dcn/edpm-post-ceph/nodeset/kustomization.yaml
@@ -148,12 +148,12 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
-      fieldPath: data.cinderAPI.customServiceConfig
+      fieldPath: data.cinder.customServiceConfig
     targets:
       - select:
           kind: OpenStackControlPlane
         fieldPaths:
-          - spec.cinder.template.cinderAPI.customServiceConfig
+          - spec.cinder.template.customServiceConfig
         options:
           create: true
   - source:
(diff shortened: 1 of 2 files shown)


abays (commented):
I'm good with this if we've tested it somewhere

dt/dcn/edpm-post-ceph/nodeset/kustomization.yaml
```diff
@@ -148,12 +148,12 @@ replacements:
...
       kind: ConfigMap
       name: service-values
-      fieldPath: data.cinderAPI.customServiceConfig
+      fieldPath: data.cinder.customServiceConfig
     targets:
       - select:
           kind: OpenStackControlPlane
         fieldPaths:
-          - spec.cinder.template.cinderAPI.customServiceConfig
+          - spec.cinder.template.cinder.customServiceConfig
```
krcmarik, line 156: I believe this should be:
 spec.cinder.template.cinder.customServiceConfig
So the generated CR would be:
```
  cinder:
    template:
      customServiceConfig: |
        [DEFAULT]
        storage_availability_zone = az0
```
And not
```
  cinder:
    template:
      cinder:
        customServiceConfig: |
          [DEFAULT]
          storage_availability_zone = az
```

fultonj:
This has passed a CI test and is ready to merge.

PR #466, opened 2024-12-17
[BGP] Replace worker-3 NAD from host-device to macvlan

eduolivares:
We implemented a solution to connect test pods running in worker-3 to the provider network using a host-device NAD. This has been working fine while only one test pod ran simultaneously.
With the new feature added to tobiko to run a tobiko-ping pod in parallel with the usual tobiko pod, attaching several pods to that NAD is necessary.
This PR modified the BGP configuration so that the mentioned host-device NAD is replaced with a macvlan NAD that allows multiple connections.


abays (approved):
/lgtm
/approve

PR #471, opened 2025-01-03
NFV network configs update

Jaganathancse:
This change is to update bridge config for NFV
deployment scenario's.


abays:
Has this been tested?

PR #474, opened 2025-01-15
WIP DNM Base definitions for uni02beta-adoption

tosky:
The net-env is a copy of the one of uni02beta (maybe there is a way to use the original one directly).


fultonj:
Closing due to no activity in 1 year. It can be re-opened if needed.

PR #481, opened 2025-01-24
Switch cinder uniquePodNames to true for DCN deployments

fmount:
This patch reverts the work done in [1], where `uniquePodNames`
was added as a `kustomization` parameter for Cinder and switched
to false due to [2].
In addition, this patch fixes the way `extraMounts` `Ceph` secrets
are accessed, and removes the not required `projected` key.

[1] https://github.com/openstack-k8s-operators/architecture/pull/436
[2] https://issues.redhat.com/browse/OSPRH-11240

Jira: https://issues.redhat.com/browse/OSPRH-11240

### DIFF
--- a/dt/dcn/control-plane/kustomization.yaml
+++ b/dt/dcn/control-plane/kustomization.yaml
@@ -91,17 +91,6 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
-      fieldPath: data.cinder.uniquePodNames
-    targets:
-      - select:
-          kind: OpenStackControlPlane
-        fieldPaths:
-          - spec.cinder.uniquePodNames
-        options:
-          create: true
-  - source:
-      kind: ConfigMap
-      name: service-values
       fieldPath: data.cinderVolumes
     targets:
       - select:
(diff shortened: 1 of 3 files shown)


tosky:
Do we want to set it explicitly to true, or remove it completely, thus relying on its default value and fully reverting https://github.com/openstack-k8s-operators/architecture/pull/436 ?
Also, a bit more details in the commit message wouldn't hurt.

fmount:
> Do we want to set it explicitly to true, or remove it completely, thus relying on its default value and fully reverting #436 ? Also, a bit more details in the commit message wouldn't hurt.

Hi @tosky, I'm ok to revert #436 entirely. I added a couple of minor changes in the way we resolve the `extraMounts` secret but they are not strictly necessary.

fmount:
> > Do we want to set it explicitly to true, or remove it completely, thus relying on its default value and fully reverting #436 ? Also, a bit more details in the commit message wouldn't hurt.
> 
> Hi @tosky, I'm ok to revert #436 entirely. I added a couple of minor changes in the way we resolve the `extraMounts` secret but they are not strictly necessary.

Done

lkuchlan:
I ran a downstream testproject that approved it.
/lgtm

PR #483, opened 2025-01-28
[Draft] remove custom ovn_emit_need_to_frag config

renjingxiao:
`custom ovn_emit_need_to_frag` config is not needed in job overrides.

Related issue: https://issues.redhat.com/browse/OSPRH-10684


abays:
@karelyatin Would you mind providing your thoughts on this?

karelyatin:
> @karelyatin Would you mind providing your thoughts on this?

@abays change in itself is fine, but i think we should wait until neutron-operator change get's shipped, /me not sure if any jobs running with released operators. The neutron-operator change merged in 18.0-fr1 and available in downstream trunk operators but not in non-trunk yet.

PR #484, opened 2025-01-29
BGP controlplane Pure L3 Spine&Leaf

eduolivares:
With this PR, BGP DT01 configures isolated L2 segments for networkers and computes.
The controlplane communications between RHOSO services running on OCP
nodes and services running on EDPM nodes is routed through the BGP
network. Spines and leaves connect the computes, networkers and OCP
nodes in different racks and route the traffic between them, as well as
the dataplane traffic to/from the Openstack VM instances.

[OSPRH-13820](https://issues.redhat.com//browse/OSPRH-13820)

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2691

### DIFF
--- a/automation/vars/bgp_dt01.yaml
+++ b/automation/vars/bgp_dt01.yaml
@@ -2,7 +2,8 @@
 vas:
   bgp_dt01:
     stages:
-      - pre_stage_run:
+      -  # stage_0
+        pre_stage_run:
           - name: Apply taint on worker-3
             type: cr
             definition:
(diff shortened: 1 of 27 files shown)


eduolivares:
recheck

abays:
Has this been tested?

eduolivares:
> Has this been tested?

@abays 
yes. I have sent you more details about the d/s verification in slack.

automation/vars/bgp_dt01.yaml
```diff
@@ -17,19 +18,65 @@ vas:
...
+          - >-
+            oc -n openstack-operators wait pod
+            -l openstack.org/operator-name=openstack
+            --for condition=Ready
+            --timeout=300s
+          - >-
+            oc -n openstack-operators wait pod
+            -l openstack.org/operator-name=infra
+            --for condition=Ready
+            --timeout=300s
```
abays, line 72: This should be unneeded as of [1] and [2] merging.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/1294
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/1292

abays:
One layout suggestion: maybe the pre-existing `examples/dt/bgp` content should be moved into a sub-directory like `bgp_dt01` has.  As it is right now, `bgp_dt01` is on the same level as `control-plane` and `edpm` dirs for `bgp`, which seems kind of awkward.  That is, maybe we could move all the things in the red box into their own sub-dir?

image

eduolivares:
> One layout suggestion: maybe the pre-existing `examples/dt/bgp` content should be moved into a sub-directory like `bgp_dt01` has. As it is right now, `bgp_dt01` is on the same level as `control-plane` and `edpm` dirs for `bgp`, which seems kind of awkward. That is, maybe we could move all the things in the red box into their own sub-dir?

That pre-existing `examples/dt/bgp` content is not useful anymore. That was the original approach for BGP, configuring all nodes on a single rack. It was replaced by bgp_dt01. I think we can remove it now.
I will do it in a separate PR, if that's ok for you.
And I will probably remove one level as well.
Something like:
```
mv examples/dt/bgp/bgp_dt01 examples/dt/
rm -rf examples/dt/bgp/
```
Better in a separate PR because there are also templates in ci-framework that need to be removed.

abays:
> > One layout suggestion: maybe the pre-existing `examples/dt/bgp` content should be moved into a sub-directory like `bgp_dt01` has. As it is right now, `bgp_dt01` is on the same level as `control-plane` and `edpm` dirs for `bgp`, which seems kind of awkward. That is, maybe we could move all the things in the red box into their own sub-dir?
> 
> That pre-existing `examples/dt/bgp` content is not useful anymore. That was the original approach for BGP, configuring all nodes on a single rack. It was replaced by bgp_dt01. I think we can remove it now. I will do it in a separate PR, if that's ok for you. And I will probably remove one level as well. Something like:
> 
> ```
> mv examples/dt/bgp/bgp_dt01 examples/dt/
> rm -rf examples/dt/bgp/
> ```
> 
> Better in a separate PR because there are also templates in ci-framework that need to be removed.

@eduolivares Ack, that sounds good to me

abays (approved):
+2

Will wait for others to concur before merging

abays:
Hmm, it seems that the explicit `/approve` and `/lgtm` tags might no longer be required.  My apologies if this merges prematurely, I didn't intend to do that.

PR #488, opened 2025-01-30
Increase data plane deployment timeout for uni06zeta

tosky:
For some reasons, the deployment of this scenario started slowing down.
As we can't easily override the timeout of the data plane deployment, bump it.


tosky:
For some reasons, the deployment of this scenario started slowing down. As we can't easily override the timeout of the data plane deployment, bump it.

abays (approved):
/lgtm
/approve

abays:
recheck-gate

PR #490, opened 2025-02-03
Decouple uni07eta network stage from control plane stage

dsariel:


dsariel:
no need in this change according to the support matrix

dsariel:
recheck

abays:
> @abays hi, where from network.tools.multus and network.network_name are coming from with relate to
> 
> https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/uni07eta/network-values/values.yaml.j2#L87?
> 
> This line is a suspect that
> 
> we have no
> 
> ```
> apiVersion: k8s.cni.cncf.io/v1
> kind: NetworkAttachmentDefinition
> metadata:
>   labels:
>     osp/net: octavia
>     osp/net-attach-def-type: standard
>   name: octavia
>   namespace: openstack
> spec:
>   config: |
>     {
>       "cniVersion": "0.3.1",
>       "name": "octavia",
>       "type": "bridge",
>       "bridge": "octbr",
>       "ipam": {
>         "type": "whereabouts",
>         "range": "172.23.0.0/24",
>         "routes":  [
>            {
>               "dst": "172.24.0.0/16",
>              "gw": "172.23.0.150"
>            }          ],
>         "range_start": "172.23.0.30",
>         "range_end": "172.23.0.70"
>       }
>     }
> ---
> ```
> 
> in logs/controller-0/ci-framework-data/artifacts/kustomize_deploy/network.yaml

@dsariel I think the base source is this [1], but that is ingested by CIFMW and then goes through further refinement/processing before it is fed to the template you mentioned [2].  I am not familiar with the internal details of that CIFMW process though.

[1] https://github.com/openstack-k8s-operators/architecture/blob/32d7703c0d522f3e0500fcabf9299fad69af9e8d/examples/dt/uni07eta/control-plane/nncp/values.yaml#L160-L183
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/uni07eta/network-values/values.yaml.j2#L87

dt/uni07eta/kustomization.yaml
```diff
@@ -344,13 +340,14 @@ replacements:
...
+# - source:
+#     kind: ConfigMap
+#     name: network-values
+#     fieldPath: data.octavia.net-attach-def
+#   targets:
+#     - select:
+#         kind: NetworkAttachmentDefinition
+#         name: octavia
+#       fieldPaths:
+#         - spec.config
```
abays, line 353: Should be able to drop this since it is moved to `dt/uni07eta/networking/kustomization.yaml` now

abays:
@dsariel Just checking in on the status of this one.  Any updates on testing?

abays:
> @dsariel Just checking in on the status of this one. Any updates on testing?

Ah, you already answered in Slack and noted testing was complete.  I will add the label.

dt/uni07eta/kustomization.yaml
abays: @dsariel Maybe we can remove this commented section since it doesn't seem to be needed?  Otherwise, given that testing is done, I think we can merge this.

abays: @dsariel If you could remove this commented section, we could merge this

automation/vars/uni07eta-adoption.yaml
```diff
@@ -0,0 +1,26 @@
+---
+vas:
+  uni07eta-adoption:
+    stages:
+      - path: examples/dt/uni07eta/control-plane/networking/nncp
```
sdatko, line 5: This file should contain basically first two stages of the original uni07eta – please add the names of the stages, so it will look exactly the same.

dsariel:
verified by https://url.corp.redhat.com/testproject

sdatko (approved):

PR #489, opened 2025-01-30
Decouple VA1 network stage from control plane stage

abays:
This removes the networking components...

1. MetalLB
2. `NetConfig`
3. Net-attach-def

...from the control plane stage of VA1 and gives them their own separate stage in between the NNCP stage and the control plane stage.  This is needed for the UniGamma adoption job, which relies upon this VA.  They want to create the NNCPs, MetalLB CRs, `NetConfig` CR and Net-attach-defs, but do NOT want to create the `OpenStackControlPlane`.  Now they can do so by creating an automation file with only two stages, one for NNCPs and the other for the other networking components.

I have tested this in my local env and the VA deployed successfully via CIFMW.

PR #498, opened 2025-02-07
Decouple uni04delta network stage from control plane stage

jirimacku:
Highly inspired by the https://github.com/openstack-k8s-operators/architecture/pull/489

This makes the networking components MetalLB NetConfig and Net-attach-def separated for the control plane stage. This allows to create an adoption job based on UniDelta architecture. The adoption job needs to deploy just the networking but not the control plane itself.


jirimacku:
Verification with regular uniDelta job passed.

PR #500, opened 2025-02-11
Update NFV VAs to use network vanity names

leifmadsen:
Reference the vanity name defined by the NNCP objects in the
L2Advertisement objects which makes the configuration more robust with
less interfaces to change.


fultonj:
This looks fine to me, it should work in theory. However, would be good to get it tested by NFV team empirically.

fultonj:
@eshulman2 @Jaganathancse are you OK with this patch?

abays (approved):
Tested by NFV DFG and succeeded

/lgtm
/approve

PR #503, opened 2025-02-17
Octavia DCN support

gthiemonge:
Enable Octavia in DCN
Network configuration for Octavia DCN support
JIRA: [OSPRH-14215](https://issues.redhat.com//browse/OSPRH-14215)

### DIFF
--- a/automation/net-env/dcn.yaml
+++ b/automation/net-env/dcn.yaml
@@ -634,6 +634,19 @@ instances:
         skip_nm: false
         trunk_parent: ctlplane
         vlan_id: 20
+      octavia:
+        interface_name: enp7s0.24
+        ip_v4: 172.23.0.10
+        is_trunk_parent: false
+        mac_addr: '52:54:00:53:17:30'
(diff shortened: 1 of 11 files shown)


krcmarik (approved):
The PR introduces dt/dcn/control-plane/nncp/* to wrap the common kustomization for nncp with octavia related bits (I believe It should go to the common one in lib/nncp not to confuse thing in specififc DCN DT but I understand that It was not approved and would require a refactor of other already created DTs this way.

abays:
> The PR introduces dt/dcn/control-plane/nncp/* to wrap the common kustomization for nncp with octavia related bits (I believe It should go to the common one in lib/nncp not to confuse thing in specififc DCN DT but I understand that It was not approved and would require a refactor of other already created DTs this way.

I think it's fine that this patch handled adding the network and its interface the way it has.  It's actually cleaner, in my opinion, than adding yet another bespoke `lib/nncp-*` directory that is then just used for one DT.

PR #504, opened 2025-02-18
Remove unused BGP configuration

eduolivares:
Configurations under the "bgp" directory implemented the first approach to test BGP on Podified setups, with all OCP nodes running on a single rack and EDPM nodes running on separate racks.
This configuration was replaced by bgp_dt01, which deploys OCP nodes and EDPM nodes splitted in 3 racks. The content of "bgp" became obsole and it is removed with this patch.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2736


abays (approved):
/lgtm
/approve

PR #513, opened 2025-02-26
Increase timeout in uni01alpha

sdatko:
My local deployments on smaller machines seem to fail with FIPS enabled scenario only due to a few minutes margin – hence I propose increasing the timeout on deployment.


sdatko:
I verified in my local downstream deployments that this improves the stability of deployments on titan machines :+1:

abays (approved):
/lgtm
/approve

abays:
recheck-gate

PR #515, opened 2025-03-04
Decouple uni04delta-ipv6 network stage from control plane stage

sauragar:
Ticket: [OSPNW-910](https://issues.redhat.com/browse/OSPNW-910)


dt/uni04delta-ipv6/networking/kustomization.yaml
```diff
@@ -0,0 +1,82 @@
...
+    unsetOnly: true
+    fieldSpecs:
+      - path: metadata/name
+        kind: Namespace
+        create: true
+
+components:
+  - ../../lib/networking/metallb
+  - ../../lib/networking/netconfig
+  - ../../lib/networking/nad
```
abays, line 22:
```suggestion
  - ../../../lib/networking/metallb
  - ../../../lib/networking/netconfig
  - ../../../lib/networking/nad
```

PR #519, opened 2025-03-13
Added name param for stages to get named stages in automation/vars for all unijobs

sauragar:
Ticket: https://issues.redhat.com/browse/OSPNW-971

### DIFF
--- a/.ci/automation-schema.yaml
+++ b/.ci/automation-schema.yaml
@@ -26,6 +26,7 @@ _hook:
   kind: str(required=False)
 ---
 _stage:
+  name: str(required=False)
   path: str()
   wait_conditions: list(str(matches='^(oc|kubectl) .* wait .*'), min=1)
   values: list(include('_values'), min=1)
--- a/automation/vars/hci-adoption.yaml
+++ b/automation/vars/hci-adoption.yaml
@@ -2,7 +2,8 @@
 vas:
   hci-adoption:
     stages:
-      - path: examples/va/hci/control-plane/networking/nncp
+      - name: nncp-configuration
+        path: examples/va/hci/control-plane/networking/nncp
         wait_conditions:
           - >-
             oc -n openstack wait nncp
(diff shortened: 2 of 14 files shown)


sdatko:
Looks great overall, but can you include also change for VA HCI scenario (https://github.com/openstack-k8s-operators/architecture/blob/main/automation/vars/hci.yaml) which is our uni03gamma scenario? :pray: 

Also, please rebase your change – I think the commit about decoupling uni05epsilon is not needed in this pull request?

sdatko:
One more thing – I see the linter is complaining about this introduced name field –> please update the .ci/automation-schema.yaml to include the additional optional field

sdatko (approved):
Looks well, was tested downstream on uni04delta-ipv6 scenario with all additional customizations for telemetry properly applied – I would say lets proceed with merging this :heart:

abays:
I wonder if we should wait for this [1] to land before merging?

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2789

sdatko:
> I wonder if we should wait for this [1] to land before merging?

Thanks for catching it. I am not sure in which case it occurred so [1] was needed, but we did not observe the problem in testing for uni04alpha or uni04alpha-ipv6; we will rerun for a few other scenarios to be fully sure before merging :pray:

krcmarik:
> > I wonder if we should wait for this [1] to land before merging?
> 
> Thanks for catching it. I am not sure in which case it occurred so [1] was needed, but we did not observe the problem in testing for uni04alpha or uni04alpha-ipv6; we will rerun for a few other scenarios to be fully sure before merging 🙏

You've probably observed the problem which [1] or [2] tries to fix but you might not realized it because It would not fail the job in the case of i.e. unialpha04* but due to the problem I believe i.e. the Ceph deployment was executed needlessly 3 times in unialpha04* jobs - you should see the "Run Deploy Ceph" three times in ansible-deploy-architecture.log. 

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2789
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/2787

sdatko:
> ... it would not fail the job in the case of i.e. unialpha04* but due to the problem I believe i.e. the Ceph deployment was executed needlessly 3 times ...

Excellent, thank you for the explanation! In such case, we are more than happy to merge the change right away, without waiting for the ongoing fix for this behavior to be merged :+1:

abays (approved):
/approve
/lgtm

PR #525, opened 2025-03-25
external netwrok list missing for networker ovs dpdk

Jaganathancse:
This change is to add external network list item in the networker ovs dpdk nodeset CR.


abays (approved):
/lgtm
/approve

PR #530, opened 2025-03-31
[BGP] Deploy frr after configure-os

eduolivares:
After https://github.com/openstack-k8s-operators/edpm-ansible/pull/897, due to the nft rules applied during the installation of frr on the EDPM nodes, ssh traffic is blocked.
That is resolved by the configure-os service, which adds nft rules to allow ssh traffic again.
Hence, the configure-os service has to be deployed before the frr service to avoid blocking ssh.


abays (approved):
/lgtm
/approve

PR #536, opened 2025-04-08
Enable octavia in uni02beta

lavraham:
We want to run the Octavia Tobiko workflow in the uni02beta job. Octavia is disabled by default in the control plane, and we need to enable Octavia to run Tobiko tests.


examples/dt/uni02beta/control-plane/networking/nncp/kustomization.yaml
```diff
@@ -18,6 +18,7 @@ transformers:
 
 components:
   - ../../../../../../lib/nncp
```
beagles, line 20: I think you can remove this line. This component is already included in the dt/uni02delta/networking/nncp component.

beagles: not in dt/uni02beta/../nncp but in examples/dt/uni02beta/control-plane/networking/nncp.
 
beagles@hardprov-fx2-16:~/architecture(main)> ./test-kustomizations.sh examples/dt/uni02beta
examples/dt/uni02beta/control-plane/networking/kustomization.yaml: OK
examples/dt/uni02beta/control-plane/networking/nncp/kustomization.yaml: OK
examples/dt/uni02beta/control-plane/kustomization.yaml: OK
examples/dt/uni02beta/kustomization.yaml: OK

beagles: Actually I had included your commits in the "wrong branch" so yeah that's still valid

abays (commented):
Looks reasonable to me, but since this affects a uni job it will need CIOps approval

PR #537, opened 2025-04-08
Add OpenStack OLM subscription configurations

sathlan:
Ability to install a specific version of the operator(s) when multiple
versions are available in the OLM catalog.

Added new kustomization files and YAML configurations for OpenStack
operator subscriptions using OLM. This includes catalog sources,
namespaces, and operator groups for openstack-operators, as well as
subscription definitions for the OpenStack components when the version
is below v1.0.7.

Kustomization facilitates version-specific configurations starting
with v1.0.3 and v1.0.6, incorporating modifications like the
elimination of the AnsibleEE operator in subsequent versions. From
version 1.0.7 onwards, only the OpenStack operator is retained.

For each overlay we define the same set of ConfigMap that reach a set
of relevant parameter in the subscriptions and catalog source
definitions.

An unexpected mechanism here is that we override the
`common/olm-subscriptions/kustomization.yaml` file from ci-framework
for version `v1.0.3` and `v1.0.6`.  This enables the dynamic selection
of the right overlay.  For version `v1.0.7` and beyond we used the
default overlay and we don't need to overwrite the `kustomization`
file.

The problems that we cannot solve with just parameters are:
- `v1.0.3` and `v1.0.6` have a different set of subscriptions to
include
- `startingCSV` parameter is "dynamic" as it's composed of the
`spec.name`.`version`, not just `version`. Thus we cannot encode in a
kustomize parameter without create our own transformer. As
transformwer are not deployed here, it would be too much work to
boostrap it.

All this disapear starting with `v1.0.7` where we have only one
subscription to set.

Resolves-Part-Of: [OSPRH-15056](https://issues.redhat.com//browse/OSPRH-15056)

### DIFF
--- a/examples/common/olm-subscriptions/README.md
+++ b/examples/common/olm-subscriptions/README.md
@@ -1,0 +1,13 @@
+# OLM subscription that are used for deploying specific versions.
+
+To deploy a specific version rather than the latest from OLM images
+containing multiple version definitions, it is necessary to explicitly
+set the `startingCSV` parameter.
+
+This kustomization setup accomplishes this for:
+- v1.0.3
+- v1.0.6: this version no longer includes the AnsibleEE operator
+- v1.0.7 and later: these versions only include the OpenStack operator
+
+Refer to the `ci_gen_kustomize_values` role README for configuration
+details.
--- a/examples/common/olm-subscriptions/kustomization.yaml
+++ b/examples/common/olm-subscriptions/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+components:
+  - ../../../lib/olm-deps
+  - ../../../lib/olm-openstack-subscriptions/overlays/default
+resources:
+  - values.yaml
--- a/examples/common/olm-subscriptions/values.yaml
+++ b/examples/common/olm-subscriptions/values.yaml
@@ -1,0 +1,17 @@
+# local-config: referenced, but not emitted by kustomize
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: olm-values
+    annotations:
+        config.kubernetes.io/local-config: 'true'
+data:
+    openstack-operator-channel: "alpha"
+    openstack-operator-image: "quay.io/openstack-k8s-operators/openstack-operator-index:latest"
+    openstack-operator-display-name: "CI Test"
+    openstack-operator-installplanapproval: "Automatic"
+    openstack-operator-publisher: "Testing Team"
+    openstack-operator-catalog-source: "openstack-operator-index"
+    openstack-operator-subscription-namespace: "openstack-operators"
+    openstack-operator-version: "latest"
--- a/lib/olm-openstack-subscriptions/base/catalogsource.yaml
+++ b/lib/olm-openstack-subscriptions/base/catalogsource.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: openstack-operator-index
+  namespace: openshift-marketplace
+  labels:
+    category: openstack-catalog
+spec:
+  image: quay.io/openstack-k8s-operators/openstack-operator-index:latest
+  sourceType: grpc
+  displayName: CI Test
+  publisher: Test Team
--- a/lib/olm-openstack-subscriptions/base/kustomization.yaml
+++ b/lib/olm-openstack-subscriptions/base/kustomization.yaml
@@ -1,0 +1,7 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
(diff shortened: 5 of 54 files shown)


examples/common/olm-subscriptions/README.md
```diff
@@ -0,0 +1,13 @@
...
+To deploy a specific version rather than the latest from OLM images
+containing multiple version definitions, it is necessary to explicitly
+set the `startingCSV` parameter.
+
+This kustomization setup accomplishes this for:
+- v1.0.3
+- v1.0.6: this version no longer includes the AnsibleEE operator
+- v1.0.7 and later: these versions only include the OpenStack operator
+
+Refer to the `ci_gen_kustomize_values` role README for configuration
```
abays, line 12: It would be good to provide a link to CIFMW if we're going to reference it, though I am a bit wary of pointing in that direction since the dependency is the other way around and we haven't explicitly mentioned CIFMW in this repo yet.

sathlan: Eventually `ci_gen_kustomize_values`  is one "hardcoded" way to interact with this (as it is for `examples/common/olm/`) and it took me some time to get it.  From a roocky point of view it makes sense to explicititely set the relation between the two, but I might miss something. Now, I get that maybe such pointer shouldn't be explicit in there.  So let me know when you've set your mind, and I'll either add a proper link, or just remove that line entirely.

examples/common/olm-subscriptions/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
sathlan, line 1: The main difference relative to to `examples/common/olm/` is that the `kustomization.yml` can be overwritten as well in order to select the right *overlay* .  I didn't find a way to make this a paramater that can be passed down to kustomize, so I  re-use the values.yml mechanism (ie overwritting file) to create the `kustomization.yaml` file as well.

lib/olm-openstack-subscriptions/overlays/v1.0.6/disable-openstack-ansibleee.yaml
sathlan, line 1: Need removal.

examples/common/olm-subscriptions/kustomization.yaml
abays: We generally assume that the various `values.yaml` `ConfigMap`s are the interface by which the user customizes the VAs/DTs.  If they're modifying the `kustomization.yaml`s, then technically they aren't using the repo as intended.  So looking at the new `examples/common/olm-subscriptions`, I'm not sure why you couldn't just add your new `values.yaml` fields/values to the existing `values.yaml` [1] and then modify the `kustomization.yaml` [2] to inject them.  Perhaps I am missing something though.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/common/olm/values.yaml
[2] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/olm-openstack/kustomization.yaml

sathlan: So the problem is twofolds here:
 - the list of subscription to create is different from version to version: 1.0.3 has ansiibleee subscription, but 1.0.6 doesn't.  That mean that the kustomization cannot include always the same list of file
 - the startingCSV value need to be `<operator subscription name>.<version>`, making it "dynamic" is it's different for every subscription.  Having overlay enable to take this into account.  But then we must choose the right overlay, hence the need to overwrite `kustomization.yaml`

Starting with v1.0.7 everything is simpler and the 'kustomization.yaml` overwrite doesn't happen and we can just use the `values.yaml` to populate the `startingCSV` of the openstack-operator to `openstack-operator.<version>` 

To summarize this is only needed to 1.0.3 and 1.0.6 and because their number of subscription is more than one and it different for each of them.

Does it make sense ?

lib/olm-openstack-subscriptions/overlays/v1.0.6/disable-openstack-ansibleee.yaml
sathlan: Removed.

examples/common/olm-subscriptions/kustomization.yaml
abays: Yes, I see what your concern is now.  I need to think about it some more, but what you've done might be the right approach for now at least.

fultonj:
> @sathlan invited me and others to a meeting in which he explained his reasoning for creating this PR this way. While it is somewhat regrettable to have to create something so specific and verbose, I agree with him that Kustomize constricts our options here. If this approach works for now, I am good with it. It does not impact the existing OLM deployment approach used by the VAs and DTs. @fultonj What do you think?

@abays I agree with you. I think we can merge.

fultonj (approved):
/approve
/lgtm

PR #539, opened 2025-04-10
Rever nncp-l3 change and switch to dedicated dir

lmiccini:


lmiccini:
ouch sorry John I'll fix this via the other bgp-l3-xl-ceph big pull, thanks for having a look.

PR #540, opened 2025-04-10
[BGP] Remove ctlplane from NNCP

eduolivares:
With BGP L3 at controlplane, OCP nodes do not need to connect to the
ctlplane directly, but they connect to EDPM nodes using BGP network
instead, so the controlplane interface can be removed from NNCP.


abays (approved):
/lgtm
/approve

PR #547, opened 2025-05-01
[OSPRH-15434] Multi-namespace VA

abays:
An architecture that represents a basic dual-namespace layout where a single RHOSO installation (which is, of course, the only option) is used to deploy two OSP clouds in separate namespaces (`openstack`, `openstack2`) within a single OCP cluster.

Depends-on: https://github.com/openstack-k8s-operators/ci-framework/pull/2941

Jira: [OSPRH-15434](https://issues.redhat.com//browse/OSPRH-15434)

### DIFF
--- a/examples/va/multi-namespace/control-plane2/kustomization.yaml
+++ b/examples/va/multi-namespace/control-plane2/kustomization.yaml
@@ -1,0 +1,82 @@
+---
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+transformers:
+  # Set namespace to OpenStack2 on all namespaced objects without a namespace
+  - |-
+    apiVersion: builtin
+    kind: NamespaceTransformer
+    metadata:
+      name: _ignored_
+      namespace: openstack2
+    setRoleBindingSubjects: none
+    unsetOnly: true
+    fieldSpecs:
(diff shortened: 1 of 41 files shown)


examples/va/multi-namespace/control-plane2/kustomization.yaml
```diff
@@ -0,0 +1,115 @@
...
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.swift.enabled
+        options:
+          create: true
+  - source:
+      kind: ConfigMap
+      name: service-values
+      fieldPath: data.telemetry.enabled
```
leifmadsen, line 86: Telemetry (metrics) needs to be disabled on all but a single control plane because it does not support multiple domains (it is not tenant aware currently).

This means Ceilometer etc should be disabled on the 2nd control plane.

Several things need to be implemented such as:

* https://issues.redhat.com/browse/RHOSSTRAT-173 (Multi-tenancy support)
* https://issues.redhat.com/browse/RHOSRFE-69 (Support for observability for multiple deployments using namespace separation)

fultonj (approved):
/approve
/lgtm

abays:
recheck-gate

PR #554, opened 2025-05-16
Added Cinder Backup in kustomization for uni02beta

sauragar:
Ticket: [OSPRH-10735](https://issues.redhat.com//browse/OSPRH-10735)

### DIFF
--- a/examples/dt/uni02beta/control-plane/service-values.yaml
+++ b/examples/dt/uni02beta/control-plane/service-values.yaml
@@ -9,6 +9,15 @@ metadata:
 
 data:
   preserveJobs: false
+
+  cinderBackup:
+    customServiceConfig: |
+      [DEFAULT]
+      backup_driver = cinder.backup.drivers.nfs.NFSBackupDriver
+      backup_share = _replaced_
+      backup_mount_point_base = _replaced_
+    replicas: 1
+
   cinderVolumes:
     ontap-nfs:
       customServiceConfig: |
--- a/dt/uni02beta/kustomization.yaml
+++ b/dt/uni02beta/kustomization.yaml
@@ -39,6 +39,28 @@ replacements:
   - source:
       kind: ConfigMap
       name: service-values
+      fieldPath: data.cinderBackup.customServiceConfig
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.cinder.template.cinderBackup.customServiceConfig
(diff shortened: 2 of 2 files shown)


examples/dt/uni02beta/control-plane/service-values.yaml
```diff
@@ -9,6 +9,15 @@ metadata:
 
 data:
   preserveJobs: false
+
+  cinderBackup:
+    customServiceConfig: |
+      [DEFAULT]
+      backup_driver = cinder.backup.drivers.nfs.NFSBackupDriver
+      backup_share = _replaced_
+      backup_mount_point_base = _replaced_
```
fultonj, line 18: Please have a better default here than _replaced_ for the backup_mount_point_base.

I used this patch to generate the new oscp yaml and the backup section looks like this

```yaml
 130   │       cinderBackup:
 131   │         customServiceConfig: |
 132   │           [DEFAULT]
 133   │           backup_driver = cinder.backup.drivers.nfs.NFSBackupDriver
 134   │           backup_share = _replaced_
 135   │           backup_mount_point_base = _replaced_
 136   │         networkAttachments:
 137   │         - storage
```

Is there a patch in ci-framework which will update the service-values so that _repaced_ gets set to an actual path?

https://docs.openstack.org/cinder/rocky/configuration/block-storage/backup/nfs-backup-driver.html

If we merge this as is, then the ci-jobs will have the value above (`_replaced_`) and I'm not sure if cinder tolerates that gracefully. Do you at least want to set it to a patch which we know will exist in the pod so that it can at least back up to a known path instead of an invalid one `_replaced_`.

fultonj (dismissed):
Please rebase and update so that backup_mount_point_base defaults to an actual path on the pod instead of `_replaced_`. If no one overrides the proposed default value it should at least be a valid path so that with this change it won't cause problems.

examples/dt/uni02beta/control-plane/service-values.yaml
sauragar: Yes @fultonj I have a patch in the ci-framework-jobs: https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/1929/diffs
I have set the value in the ci-framework-jobs and looking at other components in architecture repo I have put _replaced_ here, so is it good?
Also one more thing, should I also add the networkAttachments: storage block in the code as you mentioned here?

PR #558, opened 2025-05-21
Decouple uni02beta network from control plane and add definitions for adoption

adrianfusco:


tosky:
I see a few validation errors that should be addressed

automation/vars/uni02beta.yaml
```diff
@@ -1,9 +1,8 @@
 ---
 vas:
-  uni02beta:
+  uni02beta-adoption:
```
tosky, line 3: shouldn't this be still uni02beta: ? This is uni02beta.yaml, the variables for uni02beta-adoption are in the file above, uni02beta-adoption.yaml.

tosky (changes requested):
Also, is dt/uni02beta/networking/kustomization.yaml needed if it's empty?

sdatko:
@tosky are you sure you are on the latest? We just noticed this file missing and in last change @adrianfusco pushed it

tosky:
(long explanation, yes, I was mistaken, the issue was solved but the github UI didn't help me. And from some internal testing it looks like this works as expected)

sdatko:
I am not fully sure how it is with our permissions here, but since I can merge and it only affects uni-job, I am merging

PR #567, opened 2025-06-17
Add WatcherPassword to osp-secrets

raukadah:
Currently watcher-operator is shipped standalone operator. It is installed via custom playbook in uni03gamma job.

In order to start the deployment, we need to add watcherpassword to osp-secrets.


abays (approved):
/lgtm
/approve

PR #569, opened 2025-06-17
Update Nodeset2 to intake nova-extra configuration

jamepark4:
Nodeset2 was not utilizing the nova-extra-configuration being supplied to the job and instead was just applying the default value.  This updates nodeset2 to update the configmap for nova-extra and nova-custom if a value is supplied.


sdatko:
Cherry-picked and squashed with my change https://github.com/openstack-k8s-operators/architecture/pull/568

PR #576, opened 2025-07-02
18.0-fr3 Add designate to uni04delta

beagles:
Adds the required networking elements, enables redis instance and sets up flag for enabling designate in the dt.


beagles:
wrong base branch in error

PR #577, opened 2025-07-02
[18.0-fr3] Add designate to uni04delta

beagles:
Adds the required networking elements, enables redis instance and sets up flag for enabling designate in the dt.

Combines two PRs from main:
https://github.com/openstack-k8s-operators/architecture/pull/561
https://github.com/openstack-k8s-operators/architecture/pull/582

as 582 corrected a problem in 561


abays:
@beagles Did this pass testing?

dt/uni04delta/edpm/nodeset/kustomization.yaml
```diff
@@ -189,3 +189,75 @@ replacements:
...
+      kind: ConfigMap
+      name: service-values
+      fieldPath: data.neutron.customServiceConfig
+    targets:
+      - select:
+          kind: OpenStackControlPlane
+        fieldPaths:
+          - spec.neutron.template.customServiceConfig
+        options:
+          create: true
```
abays, line 203: I see this was removed from the control plane step and placed into a data plane step instead.  Is that because it wasn't really doing anything during the control plane step and is more properly placed here instead?

abays:
@beagles @eduolivares Just to confirm, we can close this since the branch is unused?

PR #582, opened 2025-07-07
Move designate kustomizations to nodeset for uni04delta

beagles:
Apparently uni04delta generates controlplane CR from nodeset, so moving the designate related directives to there.


eduolivares:
> @eduolivares @beagles Has this been tested?

@abays, yes. Tested with downstream testproject MR 1301.

abays (approved):
/lgtm
/approve

PR #583, opened 2025-07-07
Remove control-plane from nodeset definition

beagles:
Including the control-plane here causes kustomizations targeting the control plane to be ovewritten.


eduolivares (changes requested):
after some discussion, it seems this change is not correct and #582 should be used instead

PR #584, opened 2025-07-08
Set ovn-egress-iface=true for br-ex members

eduolivares:
In order to fix [OSPRH-17551](https://issues.redhat.com//browse/OSPRH-17551) and make BW limits properly applied to physical ports (ports from neutron VLAN and flat networks), the br-ex member interfaces need to be configured with ovn-egress-iface=true.

This PR applies the change to all uni jobs.


abays (approved):
Has been tested and verified

/lgtm
/approve

PR #588, opened 2025-07-15
[uni06zeta][neutron] remove log_dir

eduolivares:
The parameter log_dir was wrongly configured and, due to this, no neutron logs were written in uni06zeta jobs.


karelyatin (commented):
/lgtm
Also seeing some other irrelvant config which should also be cleaned up. like seeing some l3 agent/dvr specfic which not relevant with ml2/ovn.

PR #590, opened 2025-07-22
bgp_dr01: Fix NH for worker nodes

cubeek:
There was a typo where next-hops were set as addresses of BGP peer interfaces connecting to the first compute node. This patch makes the static routes for the OCP nodes being their BGP peer.

Resolves: [OSPRH-18441](https://issues.redhat.com//browse/OSPRH-18441)


abays (approved):
/lgtm
/approve

PR #591, opened 2025-07-28
Update uni07eta to involve 3 compute nodes

sdatko:
For Octavia active-standby, the testing requires
having at least 3 compute nodes, however so far
there were only 2 nodes defined in this scenario.
This commit alters the DT to match the current shape
defined in the downstream job.


abays (approved):
/lgtm

PR #595, opened 2025-08-04
[DNM] Run validate-network service one stage later

amartyasinha:
Due to some race condition, DNS service is not returning FQDN for compute nodes when validate-network pods kick in. This patch aims to start validate-network service in later stage with hope this random issue gets fixed.


slagle:
> -1
> 
> If we really think validate-network should be run later, then we should change it in EDPM itself and update our documentation.
> 
> @slagle what do you think?

We don't have a root cause of the issue that indicates such a change is needed. I agree that changing the service order here isn't right either.

amartyasinha:
> If we really think validate-network should be run later, then we should change it in EDPM itself and update our documentation.

> We don't have a root cause of the issue that indicates such a change is needed. I agree that changing the service order here isn't right either.
@evallesp WDYT about this? It is true that we don't have a root cause. This issue is happening randomly and we are trying to avoid it by changing the order of services. What should be the next step as reordering services without root cause is discouraged.

amartyasinha:
Unfortunately, logs have been rotated. And since it's occurring rarely, it
is difficult to reproduce it.

karelyatin:
<< Unfortunately, logs have been rotated. And since it's occurring rarely, it
is difficult to reproduce it.
ack then let's  hold this and edpm-ansible PR as slagle pointed out we should check the issue once we have logs and/or reproducer

PR #597, opened 2025-08-06
Add second nodeset to hci VA

ciecierski:
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3186


fultonj:
maybe you could do a separate DT instead of doing this to va1?

PR #598, opened 2025-08-08
Drop unused files

sdatko:
That files are no longer used, were replaced with
corresponding ones in `lib/control-plane/base/` directory,
but during migration the originals were not cleaned.
This commit removes them.


abays (approved):
/lgtm
/approve

PR #599, opened 2025-08-08
Allow setting TLS-related options

sdatko:
This would allow us to define settings, such as `caBundleSecretName`,
which are useful in case of deployments that require establishing
connections with internal/local services, for example that utilize
self-signed certificates.


kstrenkova:
@abays Hi! Could I ask you for an approval? The bot wants you: `Once this PR has been reviewed and has the lgtm label, please ask for approval from [abays](https://github.com/abays)`

PR #603, opened 2025-08-18
Add watcher deployment integrated on openstack-operator

morenod:
Adding watcher to be deployed by openstack-operator with kustomized config


examples/va/hci/deployment/values.yaml
```diff
@@ -17,3 +17,4 @@ data:
     - libvirt
     - nova-custom-ceph
     - telemetry
+    - watcher
```
raukadah, line 20: It is not needed as watcher service is only deployed on control plane.

PR #610, opened 2025-08-28
Add some missing Kustomize APIVersion and Kind fields

abays:
Found a few `kustomization.yaml`s that were missing `APIVersion` and `Kind`.  They still work without them, but let's make sure they're canonically correct.

Co-authored-by: Claude [claude@anthropic.com](mailto:claude@anthropic.com)


fultonj (approved):
/approve
/lgtm

PR #621, opened 2025-09-18
Add databaseInstance for manila in ctlplane template

stuggi:
There seems to be a bug in the current controller-gen version used, which resulted in the databaseInstance in manila not to be a required paramater.
With the bump controller-gen happening as part of the golang-bump work the databaseInstance gets added as a required parameter [1]. This adds it to the ctlplane template.

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/462#discussion_r2293947364


fmount (approved):
/lgtm

stuggi:
in running a test the ctlplane deployed ok with this
```
$ oc get osctlplane
NAME           STATUS   MESSAGE
controlplane   True     Setup complete
```

PR #624, opened 2025-09-22
[BGP] Run bootstrap before download-cache

eduolivares:
The BGP jobs were the only ones running download-cache EDPM service before bootstrap. This worked well until [1].
The bootstrap service has to run first, in order to generate the bootc fact.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/1018


eduolivares:
recheck

PR #625, opened 2025-09-25
[BGP] Remove NNCP for master nodes

eduolivares:
In BGP DTs, NNCP only makes sense for OCP workers because no RHOSO service is deployed on OCP masters.


fultonj:
This is still being tested

ccamposr (approved):

evallesp (approved):
LGTM

PR #626, opened 2025-09-29
Add storagemgmt NAD to uni05epsilon DT

ashu-011:
This commit adds storagemgmt NetworkAttachments Definition to uni05epsilon DT

### DIFF
--- a/examples/dt/uni05epsilon/control-plane/networking/nncp/kustomization.yaml
+++ b/examples/dt/uni05epsilon/control-plane/networking/nncp/kustomization.yaml
@@ -21,3 +21,109 @@ components:
 
 resources:
   - values.yaml
+
+patches:
+  - target:
+      kind: NodeNetworkConfigurationPolicy
+      name: "master-.*"
+    patch: |-
+      - op: add
+        path: /spec/desiredState/interfaces/-
+        value:
+          description: storagemgmt vlan host interface
+          name: storagemgmt
+          state: up
+          type: vlan
+          mtu: _mtu_replaced_
+          ipv4:
+            address:
+            - ip: _ip_replaced_
+              prefix-length: _prefix_replaced_
+            dhcp: false
+            enabled: true
+          ipv6:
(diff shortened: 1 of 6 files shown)


fultonj:
The author was "told the standard approach was to always use storagemgmt [for swift] in pre 18 release" so I am assuming swift is actually using it. I was not aware of that. I don't see why it wouldn't be possible in theory.

  https://github.com/openstack-k8s-operators/data-plane-adoption/pull/1063/files

fultonj:
Let's get some testing results before merging

examples/dt/uni05epsilon/control-plane/networking/nncp/kustomization.yaml
```diff
@@ -21,3 +21,117 @@ components:
...
+  - source:
+      kind: ConfigMap
+      name: network-values
+      fieldPath: data.storagemgmt.vlan
+    targets:
+      - select:
+          kind: NodeNetworkConfigurationPolicy
+          name: master-0
+        fieldPaths:
+          - spec.desiredState.interfaces.[name=storagemgmt].vlan.id
```
abays, line 93: You can simplify this and remove the need for separate ones for master-1 and master-2, since the VLAN ID is the same across all nodes.
```suggestion
  - source:
      kind: ConfigMap
      name: network-values
      fieldPath: data.storagemgmt.vlan
    targets:
      - select:
          kind: NodeNetworkConfigurationPolicy
          name: .*
        fieldPaths:
          - spec.desiredState.interfaces.[name=storagemgmt].vlan.id
```

bogdando:
We've got a [1st pass](https://gitlab.cee.redhat.com/ci-framework/ci-framework-testproject/-/merge_requests/1539) for uni05epsilon adoption!

@ashu-011 we need another one testproject to also verify that uni05epsilon greenfield jov

otherwise, LGTM!

abays (approved):
/lgtm
/approve

PR #724, opened 2026-03-19
Add nova05epsilon for DCN with BM SNO+EDPM

bogdando:
Based on hci, dcn and nova04delta DTs, but
adapted for SNO. Octavia is removed.

Jira: [OSPRH-27841](https://redhat.atlassian.net/browse/OSPRH-27841)
Generated-by: claude-4.6-opus-high

### DIFF
--- a/automation/vars/nova05epsilon.yaml
+++ b/automation/vars/nova05epsilon.yaml
@@ -1,0 +1,134 @@
+---
+vas:
+  nova05epsilon:
+    stages:
+      - name: nncp-configuration
+        path: examples/dt/nova/nova05epsilon/control-plane/networking/nncp
+        wait_conditions:
+          - >-
+            oc -n openstack wait nncp
+            -l osp/nncm-config-type=standard
+            --for jsonpath='{.status.conditions[0].reason}'=SuccessfullyConfigured
+            --timeout=5m
+        values:
+          - name: network-values
+            src_file: values.yaml
+        build_output: nncp.yaml
+
+      - name: network-configuration
+        path: examples/dt/nova/nova05epsilon/control-plane/networking
+        wait_conditions:
+          - >-
+            oc -n metallb-system wait pod
+            -l app=metallb -l component=speaker
+            --for condition=Ready
+            --timeout=5m
+        values:
+          - name: network-values
+            src_file: nncp/values.yaml
+        build_output: network.yaml
+
+      - name: control-plane
+        path: examples/dt/nova/nova05epsilon/control-plane
+        wait_conditions:
+          - >-
+            oc -n openstack wait osctlplane controlplane
+            --for condition=OpenStackControlPlaneDNSReadyCondition
+            --timeout=20m
+        values:
+          - name: network-values
+            src_file: networking/nncp/values.yaml
+          - name: service-values
+            src_file: service-values.yaml
+        build_output: control-plane.yaml
+
+      - name: dns-configuration
+        path: examples/dt/nova/nova05epsilon/control-plane/networking/dns
+        wait_conditions:
+          - >-
+            oc -n openshift-dns wait dns.operator/default
+            --for condition=Available
+            --timeout=20m
+        values:
+          - name: dns-values
+            src_file: values.yaml
+        build_output: dns.yaml
+
+      - name: baremetalhosts-configuration
+        path: examples/dt/nova/nova05epsilon/edpm/baremetalhosts
+        wait_conditions:
+          - >-
+            oc -n openstack wait baremetalhosts.metal3.io edpm-compute-0
+            --for=jsonpath='{.status.provisioning.state}'=available
+            --timeout=20m
+          - >-
+            oc -n openstack wait osctlplane controlplane
+            --for condition=Ready
+            --timeout=20m
+        values:
+          - name: baremetalhost-values
+            src_file: values.yaml
+        build_output: baremetalhosts.yaml
+
+      - name: edpm-nodeset-pre-ceph
+        path: examples/dt/nova/nova05epsilon/edpm-pre-ceph/nodeset
+        wait_conditions:
+          - >-
+            oc -n openstack wait osdpns
+            gpu-computes-edpm
+            --for condition=SetupReady
+            --timeout=10m
+        values:
+          - name: edpm-nodeset-values
+            src_file: values.yaml
+        build_output: nodeset-pre-ceph.yaml
+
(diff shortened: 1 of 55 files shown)


bogdando:
Tested downstream

zuul.d/validations.yaml
```diff
@@ -305,6 +305,22 @@
...
+- job:
+    files:
+    - examples/dt/nova/nova05epsilon
+    - examples/dt/nova/nova05epsilon/control-plane
+    - examples/dt/nova/nova05epsilon/control-plane/networking
+    - examples/dt/nova/nova05epsilon/control-plane/networking/dns
+    - examples/dt/nova/nova05epsilon/control-plane/networking/nncp
+    - examples/dt/nova/nova05epsilon/deployment
+    - examples/dt/nova/nova05epsilon/edpm-pre-ceph/deployment
+    - examples/dt/nova/nova05epsilon/edpm-pre-ceph/nodeset
```
abays, line 317: Is there any particular reason that `edpm-post-ceph` stuff is excluded from this list?

bogdando: nicely spotted, thanks!

bogdando: The edpm-post-ceph component in dt/nova/nova05epsilon/edpm-post-ceph/ is consumed by the top-level example kustomization and the deployment/ kustomization. No additional example directories are needed — this is exactly how DCN, VA/HCI, and NFV/HCI all work.

abays (commented):
/lgtm

Will defer to @fultonj for approval

automation/vars/nova05epsilon.yaml
```diff
@@ -0,0 +1,135 @@
...
+          - name: network-values
+            src_file: nncp/values.yaml
+        build_output: network.yaml
+
+      - name: control-plane
+        path: examples/dt/nova/nova05epsilon/control-plane
+        wait_conditions:
+          - >-
+            oc -n openstack wait osctlplane controlplane
+            --for condition=OpenStackControlPlaneDNSReadyCondition
```
fultonj, line 36: I see it's waiting for `OpenStackControlPlaneDNSReadyCondition`. 

Do you only need to  DNS service to be ready at this stage?

The `nova05epsilon/examples/dt/nova/nova05epsilon/control-plane.md` contained in this PR has:

```
oc wait osctlplane controlplane --for condition=Ready --timeout=600s
```

Should these be consistent?

fultonj (commented):
This DT is really good and I don't think we need many changes in the YAML (aside from a question I raised below about the DNS Ready condition).

The documentation in the markdown is very thorough but I'd like to request some additions because of DCN storage requirements.

This DT has the potential to be extended with more compute AZs for DCN, but I'm concerned it might mislead people to break DCN requirements when storage is used as the markdown is currently written.

If the HCI compute node in the nova05epsilon DT were hosted at a remote site (as implied by calling this DCN), then there would not be any local storage for Glance running on the control plane. The DCN DT avoids this problem by requiring one of the Ceph clusters to be local.

If it's important that this be a DCN DT (instead of just calling it "Nova GPU Passthrough (VFIO) with CephHCI over routed networking"), then could we be more explicit about its requirements when storage is used?

I see some text has been written about if multiple ceph clusters are used in a DCN scenario. 

Do you want to add some text like the following to architecture/examples/dt/nova/nova05epsilon/README.md to cover how it could be extended for use in DCN with storage?

```
## Extending to a Full DCN Deployment

This topology deploys a single availability zone with computes on a
remote L3 subnet. The multi-subnet networking (subnet1 for SNO,
subnet2 for EDPM computes) provides the routed-networking foundation
for DCN, but additional changes are required to make this a multi-AZ 
distributed compute deployment with stroage. Use [dt/dcn](../../dcn/README.md)
as a reference.

### Glance and Local Ceph Storage

In this topology, it is implied that Ceph runs on remote EDPM compute
nodes (CephHCI), but Glance runs on the control plane at another site.
This means Glance's RBD backend is accessing a Ceph cluster over the WAN,
which adds latency to Glance API image operations. In a DCN deployment with
storage, at least one Ceph cluster must be local to the control plane
so that the central Glance instance has low-latency access to its default
RBD store. The DCN DT achieves this by designating az0 as the central site
with its own local Ceph cluster and configuring Glance az0 as type `split`
with `default_backend = az0`. Edge sites use Glance type `edge` and
replicate images from the central store on demand. This latency requirement
is pronounced when Cinder is used.

### Per-AZ Ceph Configuration

In `values.yaml`, switch from plain Ceph filenames (`ceph.conf`,
`ceph.client.openstack.keyring`) to az-prefixed filenames
(`az0.conf`, `az0.client.openstack.keyring`, `az1.conf`, etc.) so
that each site's Ceph cluster is independently addressable. Update
the Nova Ceph config (`nova.ceph.conf`) to reference the correct
az-prefixed conf file for each site.

### Per-AZ Glance

Replace the single Glance backend with per-AZ `glanceAPIs` entries
in `service-values.yaml`. The central site should be type `split`
with backends for all AZs; edge sites should be type `edge` with
backends for their own AZ plus the central AZ (for copy-on-write
image import).

### OVN Availability Zones

Set `edpm_ovn_availability_zones` in the nodeset values for each
site instead of leaving it empty. This enables OVN chassis-level AZ
awareness for network scheduling.

### Additional Networking Subnets

Add a subnet per site for each network (ctlplane, internalapi,
storage, tenant) in the network values. The current topology
defines subnet1 and subnet2; a full DCN deployment with N sites
needs N+1 subnets (one for the control plane site, one per edge
site).

### Per-Site Dataplane Stages

The dataplane stages (pre-ceph nodeset, pre-ceph deployment, Ceph
install, post-ceph nodeset, post-ceph deployment) must be repeated
for each DCN site with site-specific values. Each site's computes
should reference the appropriate cell, subnet, and AZ.
```

automation/vars/nova05epsilon.yaml
bogdando: nova05epsilon has a separate dns-configuration stage (stage 3 in cifmw jobs) right after the control-plane stage. The idea is: wait only for DNS at stage 2, then apply the dns operator config at stage 3, then proceed with BMH at stage 4 (which does wait for condition=Ready).

bogdando:
> architecture/examples/dt/nova/nova05epsilon/README.md

@fultonj updated

fultonj (approved):
/approve
/lgtm

PR #735, opened 2026-04-08
[uni04delta-ipv6] set mac binding age threshold to 0

averdagu:
Due to a regression in OVN, we're observing in this job that the snat ping is failing, this is related to the mac bindings, setting this to 0 as a workarround until issue is fixed either on neutron or OVN.

Related: [OSPCIX-1304](https://redhat.atlassian.net/browse/OSPCIX-1304)


eduolivares (approved):

ralonsoh (approved):

eduolivares:
I have added the "hold" label, since we tested this PR again and apparently it doesn't resolve the issue 100% times.

PR #738, opened 2026-04-12
[SKMO] Enable Cinder Backup with Swift backend

lkuchlan:
Add Cinder Backup service configuration for va-multi-namespace-skmo Swift is deployed in the central region (regionOne) and shared across both regions via Keystone service catalog. Both regions' cinder-backup services use single_user authentication to access Swift

Changes:
- control-plane: Enable Swift service and add cinderBackup config
- control-plane2: Add cinderBackup configuration for leaf region


fultonj (commented):
/lgtm

Want to get approval from @vakwetu before merging

vakwetu (approved):

abays (approved):
/lgtm
/approve

PR #740, opened 2026-04-13
skmo: use post-deploy federation controlplane hook

vakwetu:
Switch the Stage 5 post_stage_run from federation-controlplane-config.yml to federation-controlplane-config-postdeploy.yml.

The post-deploy hook is designed for pipelines where the OpenStackControlPlane CR already exists.  It reads the live OSCP to determine the CA bundle secret name, preserves any existing secret data, and merges in the Keycloak CA cert before patching caBundleSecretName when not already set.

The original federation-controlplane-config.yml (pre-deploy hook) is intended for the component pipeline where the OSCP does not yet exist at hook execution time.

Relates-to: [OSPCIX-1321](https://redhat.atlassian.net/browse/OSPCIX-1321)
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3847

### DIFF
--- a/automation/vars/multi-namespace-skmo.yaml
+++ b/automation/vars/multi-namespace-skmo.yaml
@@ -100,11 +100,11 @@ vas:
             src_file: ../../multi-namespace/control-plane/networking/nncp/values.yaml
         build_output: ../control-plane.yaml
         post_stage_run:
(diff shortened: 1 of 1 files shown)


evallesp (approved):
LGTM

danpawlik (approved):

PR #743, opened 2026-04-17
Add ordering prefixes to BGP pre_stage_run hooks

abays:
When a stage's pre_stage_run or post_stage_run contains more than one hook, each hook name must include a numeric ordering prefix (e.g. "01", "02") to ensure they execute in a deterministic order.  Even though the hooks are a list, the ordering actually depends on the alphabetical sort of the hooks names, which is fragile and not guaranteed to match the intended sequence.

From ci-framework/roles/run_hook/tasks/main.yml around line 74:                                                                                                                                                                                                          
     
```             
  _hooks: >-                                                                                                                                                                                                                                                                                
    {{            
      (                                                                                                                                                                                                                                                                                     
        _single_hooks +                                                                                                                                                                                                                                                                     
        (hooks | default([], true)) +                                                                                                                                                                                                                                                       
        _list_hooks
      ) | sort(attribute='name')
    }}                                                                                                                                                                                                                                                                                      
```

The `| sort(attribute='name')` Jinja2 filter explicitly reorders all hooks alphabetically before the loop executes them.

Made-with: Cursor

### DIFF
--- a/automation/vars/bgp-l3-xl.yaml
+++ b/automation/vars/bgp-l3-xl.yaml
@@ -3,7 +3,7 @@ vas:
   bgp-l3-xl:
     stages:
       - pre_stage_run:
-          - name: Apply taint on worker-9
+          - name: 01 Apply taint on worker-9
             type: cr
             definition:
               spec:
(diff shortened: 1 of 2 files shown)


lmiccini (approved):
this is karma at its best!

PR #744, opened 2026-04-17
Add CI validation for hook ordering prefixes

abays:
CIFMW's run_hook module sorts hooks alphabetically by name before executing them. When a pre_stage_run or post_stage_run stage has two or more hooks, this alphabetical sorting can silently alter the intended execution order unless each hook name carries a zero-padded numeric prefix (e.g. "01 Install operator", "02 Create site").

Add .ci/validate-hook-ordering.py to enforce that:
- Every hook in a multi-hook stage has a numeric prefix.
- Prefixes are sequential with no gaps or duplicates.

Wire the script into the GitHub Actions workflow
(.github/workflows/automation-schema.yaml) by renaming the files_exist job to validate_automation and adding a step to run the new validator.

Co-author: Claude Opus 4.6

### DIFF
--- a/.ci/validate-hook-ordering.py
+++ b/.ci/validate-hook-ordering.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+"""Validate numeric prefix ordering on pre/post stage hooks.
+
+When a stage has multiple hooks, run_hook sorts them alphabetically by name.
+To guarantee execution order, each hook name must start with a zero-padded
+numeric prefix (e.g. "01 Install operator", "02 Create site").
+
+This script checks:
+  - If a stage has >1 hook, every hook name starts with a numeric prefix.
+  - Prefixes are sequential (01, 02, 03, ...) with no gaps or duplicates.
+  - If a stage has exactly 1 hook, a prefix is allowed but not required.
+"""
+
+import pathlib
+import re
+import sys
+
+import yaml
+
+PREFIX_RE = re.compile(r'^(\d+)\s+')
+
+def validate_hooks(hooks, stage_label):
+    """Validate a list of hooks. Returns list of error strings."""
+    errors = []
+    if not hooks or len(hooks) < 2:
+        return errors
+
+    prefixes = []
+    for hook in hooks:
+        name = hook.get('name', '<unnamed>')
+        m = PREFIX_RE.match(name)
+        if not m:
+            errors.append(
+                f'{stage_label}: hook "{name}" is missing a numeric prefix '
+                f'(required when stage has {len(hooks)} hooks)'
+            )
+        else:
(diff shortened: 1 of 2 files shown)


.ci/validate-hook-ordering.py
```diff
@@ -0,0 +1,102 @@
...
+
+    prefixes = []
+    for hook in hooks:
+        name = hook.get('name', '<unnamed>')
+        m = PREFIX_RE.match(name)
+        if not m:
+            errors.append(
+                f'{stage_label}: hook "{name}" is missing a numeric prefix '
+                f'(required when stage has {len(hooks)} hooks)'
+            )
```
fultonj, line 36: Nit: Claude code suggested that the validation should enforce zero-padding:

Without zero-padding, a stage with 10+ hooks would silently execute in the wrong order:
```
"1 foo"   ← sorts first alphabetically
"10 bar"  ← sorts second (before "2")
"2 baz"   ← sorts third
```
The validator without the patch would see integers [1, 2, 10] — not consecutive — and correctly reject that. But [1, 2, 3, ..., 9, 10] is consecutive, so it would pass, even though alphabetically "10 ..." sorts before "2 ...".

With zero-padding, the alphabetic sort and the intended numeric order always agree:
```
"01 foo"
"02 baz"
"10 bar"  ← now sorts correctly after 09
```
In practice the current repo caps at 3 hooks per stage, so it won't be triggered today. But the validator is meant to protect future contributors who don't know this gotcha.

Here's the suggested change:
```
  37   │         else:
  38 + │             if not m.group(1).startswith('0') and len(m.group(1)) < 2:
  39 + │                 errors.append(
  40 + │                     f'{stage_label}: hook "{name}" prefix must be zero-padded '
  41 + │                     f'(e.g. "01" not "1")'
  42 + │                 )
  43   │             prefixes.append((int(m.group(1)), name))
```

fultonj (commented):
Thanks Andrew. This looks good. I was about to merge but wanted to point out something claude code discovered in case you want to add it first.

.ci/validate-hook-ordering.py
abays: Thank you John!  I've pushed the suggested change.

fultonj (approved):
/lgtm
/approve

abays:
recheck-gate

abays:
Just a rebase, as was required.  Re-adding LGTM label.

abays:
recheck-gate

abays:
recheck-gate

autkin-redhat:
recheck-gate

autkin-redhat (approved):

abays:
Merging manually due to CI auto-merge issue.

PR #746, opened 2026-04-20
Add hci-ironic VA variant for Ironic-provisioned HCI deployments

rebtoor:
Add a new validated architecture variant (hci-ironic) that deploys
the VA-HCI scenario with all compute nodes provisioned via Ironic
using a configurable baremetalSetTemplate.osImage, enabling validation
of edpm-hardened-uefi qcow2 images through a complete deployment cycle.

New reusable component:
- lib/dataplane/nodeset-baremetal -- kustomize Component that maps
  preProvisioned and baremetalSetTemplate from the values ConfigMap
  into the OpenStackDataPlaneNodeSet spec.

New VA variant (va/hci-ironic):
- Identical to va/hci except the edpm-pre-ceph nodeset stage includes
  nodeset-baremetal alongside the standard nodeset component.
- All other stages (NNCP, networking, control-plane, deployments,
  post-ceph, Ceph bootstrap hook) reuse the existing va/hci paths.
- SetupReady timeout increased to 30m to account for Ironic
  provisioning time.

Documentation:
- Added README and dataplane-pre-ceph guide for the hci-ironic variant.
- Updated the standard va/hci README with a Variants section pointing
  to hci-ironic for Ironic-provisioned deployments.

The standard va/hci is untouched -- existing pre-provisioned
deployments are not affected.

Closes: [ANVIL-108](https://redhat.atlassian.net/browse/ANVIL-108)

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/examples/va/hci-ironic/edpm-pre-ceph/nodeset/values.yaml
+++ b/examples/va/hci-ironic/edpm-pre-ceph/nodeset/values.yaml
@@ -1,0 +1,192 @@
+# yamllint disable rule:line-length
+# local-config: referenced, but not emitted by kustomize
+#
+# Ironic-provisioned variant: includes preProvisioned and
+# baremetalSetTemplate for Ironic provisioning via Metal3.
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: edpm-nodeset-values
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  preProvisioned: false
+  baremetalSetTemplate:
+    # CHANGEME - qcow2 image name from the edpm-hardened-uefi container
+    osImage: edpm-hardened-uefi.qcow2
+    # set to 'metadata' if redeploying Ceph to ensure clean disks for OSDs
+    automatedCleaningMode: metadata
+    cloudUserName: cloud-admin
+    bmhLabelSelector:
+      app: openstack
+    bmhNamespace: openshift-machine-api
+  ssh_keys:
+    # Authorized keys that will have access to the dataplane computes via SSH
+    authorized: CHANGEME
+    # The private key that will have access to the dataplane computes via SSH
+    private: CHANGEME2
+    # The public key that will have access to the dataplane computes via SSH
+    public: CHANGEME3
+  nodeset:
+    ansible:
+      ansibleUser: cloud-admin
+      ansiblePort: 22
+      ansibleVars:
+        timesync_ntp_servers:
+          - hostname: pool.ntp.org
+        # CHANGEME -- see https://access.redhat.com/solutions/253273
+        # edpm_bootstrap_command: |
+        #       subscription-manager register --username <subscription_manager_username> \
+        #         --password <subscription_manager_password>
+        #       podman login -u <registry_username> -p <registry_password> registry.redhat.io
+        edpm_network_config_hide_sensitive_logs: false
+        edpm_network_config_os_net_config_mappings:
+          edpm-compute-0:
+            nic2: 6a:fe:54:3f:8a:02  # CHANGEME - should connect to same network as control plane enp6s0
+          edpm-compute-1:
+            nic2: 6b:fe:54:3f:8a:02  # CHANGEME - should connect to same network as control plane enp6s0
+          edpm-compute-2:
+            nic2: 6c:fe:54:3f:8a:02  # CHANGEME - should connect to same network as control plane enp6s0
+        edpm_network_config_template: |
+          ---
+          {% set mtu_list = [ctlplane_mtu] %}
+          {% for network in nodeset_networks %}
+          {% set _ = mtu_list.append(lookup('vars', networks_lower[network] ~ '_mtu')) %}
+          {%- endfor %}
+          {% set min_viable_mtu = mtu_list | max %}
+          network_config:
+          - type: interface
+            name: nic1
+            use_dhcp: true
+            mtu: {{ min_viable_mtu }}
+          - type: ovs_bridge
+            name: {{ neutron_physical_bridge_name }}
+            mtu: {{ min_viable_mtu }}
+            use_dhcp: false
+            dns_servers: {{ ctlplane_dns_nameservers }}
+            domain: {{ dns_search_domains }}
+            addresses:
+            - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_cidr }}
+            routes: {{ ctlplane_host_routes }}
+            members:
+            - type: interface
+              name: nic2
+              mtu: {{ min_viable_mtu }}
+              # force the MAC address of the bridge to this interface
+              primary: true
+          {% for network in nodeset_networks %}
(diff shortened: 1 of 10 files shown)


rebtoor:
recheck

fultonj (commented):
Hey Roberto,

This looks really good. Would you please consider two changes?

1. Add Markdown files

Please add markdown files which describe how to build this VA. 
For example someone without ci-framework can browse to this URL:

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/README.md

and then build the CRs by following the directions and running kustomize.

The new markdown files should make it clear what this alternative to the hci VA are.

I suggest also updating the README for va/hci to state something along the lines of:

```
The steps in [Configuring and deploying the pre-Ceph dataplane](examples/va/hci/dataplane-pre-ceph.md)
assume that the compute nodes have been pre-provisioned. If you wish to pre-provision these nodes with
Ironic see ...
```

Consider one set of directions with two possible methods of deploying HCI.
One with pre-provisioned EDPM nodes and one without.

2. Naming: Baremetal vs Ironic

I have a small concern that calling this "baremetal" implies that the original HCI is only for running in VMs which is not the case.
Would it be difficult in the naming for "baremetal" be replaced by "ironic" since the difference here is really about how the EDPM nodes were deployed?

Other than that this looks really good to me.

I built the CRs for both hci and hci-baremetal and all CRs are the same except nodeset-pre-ceph.yaml. Which is exactly what I would expect.

```diff
--- hci/nodeset-pre-ceph.yaml   2026-04-22 13:58:44.208210525 -0400
+++ hci-baremetal/nodeset-pre-ceph.yaml 2026-04-22 13:57:26.240705158 -0400
@@ -34,6 +34,13 @@
   name: openstack-edpm
   namespace: openstack
 spec:
+  baremetalSetTemplate:
+    automatedCleaningMode: disabled
+    bmhLabelSelector:
+      app: openstack
+    bmhNamespace: openshift-machine-api
+    cloudUserName: cloud-admin
+    osImage: edpm-hardened-uefi.qcow2
   env:
   - name: ANSIBLE_FORCE_COLOR
     value: "True"
@@ -175,7 +182,7 @@
         subnetName: subnet1
       - name: tenant
         subnetName: subnet1
-  preProvisioned: true
+  preProvisioned: false
   services:
   - bootstrap
   - configure-network
```

rebtoor:
Hey John! Thanks for your careful review!

I've addressed both of your requests, I agree with you that ironic > baremetal has much more sense in this context, 'cause the latter would have been misleading so i renamed the scenario (and the downstream job as well).

examples/va/hci/control-plane/control-plane
```diff
@@ -0,0 +1 @@
+../hci/control-plane
```
fultonj: I assume this was an accidental symlink. Want to remove it in the next patchset?
```
[johfulto@laptop control-plane{baremetal}]$ ll
total 12K
-rw-r--r--. 1 johfulto johfulto 419 Jan 13 16:31 kustomization.yaml
drwxr-xr-x. 1 johfulto johfulto  44 Jan 13 16:31 networking/
-rw-r--r--. 1 johfulto johfulto 345 Apr 13 11:42 service-values.yaml
lrwxrwxrwx. 1 johfulto johfulto  20 Apr 23 14:47 control-plane -> ../hci/control-plane
[johfulto@laptop control-plane{baremetal}]$ 
```

examples/va/hci-ironic/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+# This is the kustomization for the FINAL step, edpm-post-ceph
+# (hci-ironic variant: references hci-ironic pre-ceph values)
+---
```
fultonj: `examples/va/hci-ironic/kustomization.yaml` as it is right now is  broken [1]. 

The current README doesn't direct users to use it and `automation/vars/hci-ironic.yaml` does not use it either. Thus, I think it could just be removed.

[1]
```
[johfulto@laptop hci-ironic{baremetal}]$ kustomize build .
Error: accumulating resources: accumulation err='accumulating resources from 'control-plane/networking/nncp/values.yaml': evalsymlink failure on '/home/johfulto/claude/review/architecture/examples/va/hci-ironic/control-plane/networking/nncp/values.yaml' : lstat /home/johfulto/claude/review/architecture/examples/va/hci-ironic/control-plane: no such file or directory': must build at directory: not a valid directory: evalsymlink failure on '/home/johfulto/claude/review/architecture/examples/va/hci-ironic/control-plane/networking/nncp/values.yaml' : lstat /home/johfulto/claude/review/architecture/examples/va/hci-ironic/control-plane: no such file or directory
[johfulto@laptop hci-ironic{baremetal}]$
```

examples/va/hci/README.md
```diff
@@ -29,6 +29,13 @@ This is a collection of CR templates that represent a validated Red Hat OpenStac
 
 5. Between stages 3 and 4, _it is assumed that the user installs Ceph on the 3 OSP compute nodes._  OpenStack K8S CRDs do not provide a way to install Ceph via any sort of combination of CRs.
 
+## Variants
```
fultonj, line 32: nice!

fultonj (commented):
Thanks @rebtoor ! 

Just two small changes and I'm good to go.

examples/va/hci-ironic/edpm-pre-ceph/nodeset/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+metadata:
+  name: edpm-nodeset-values
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  preProvisioned: false
+  baremetalSetTemplate:
+    # CHANGEME - qcow2 image name from the edpm-hardened-uefi container
+    osImage: edpm-hardened-uefi.qcow2
+    automatedCleaningMode: disabled
```
fultonj: Would you please add the following comment above `automatedCleaningMode` ?
```
# set to 'metadata' if redeploying Ceph to ensure clean disks for OSDs 
```
Or perhaps change `disabled` to `metadata`?

That would be consistent with what we used to recommend with TripleO:

https://docs.openstack.org/project-deploy-guide/tripleo-docs/wallaby/features/cephadm.html#prerequisite-ensure-disks-are-clean

When deploying Ceph, OSDs will not be created unless the disk is factory clean. This subtlety presents itself when people redeploy (since it's hard to get everything right the first time) and the Ceph install fails. Support used to get lots of calls about OSDs not getting created because the reader didn't know this.

I confirmed `metadata` is the correct setting as per:

https://book.metal3.io/capm3/automated_cleaning.html

fultonj (approved):
/approve
/lgtm

evallesp (approved):

rebtoor:
@fultonj @abays can we merge this? :)

arxcruz (approved):
/approve
/lgtm

PR #748, opened 2026-04-27
uni05epsilon: extract shared controlplane replacements to avoid duplication

lmiccini:
The edpm-post-ceph nodeset kustomization was duplicating all 
epsilon-specific controlplane replacements from the control-plane
stage (galera-cell2, rabbitmq-cell2, nova-cell2). The 
rabbitmq-notifications replacement was missed, causing the operator
to delete rabbitmq-notifications when the edpm-post-ceph stage
was applied.

Extract the shared replacements into a new controlplane-overrides
component that both the control-plane stage and edpm-post-ceph stage
include. This prevents future drift when new replacements are added.


belolipa (approved):
/lgtm
/approve

abays (approved):
/lgtm
/approve

PR #749, opened 2026-04-28
Add some more detail to AGENTS.md

abays:
Just a bit of additional detail in two sections of `AGENTS.md`, and then some guidance on testing in CIFMW context.


rebtoor (approved):
/lgtm

PR #751, opened 2026-04-29
DNM Revert "Switch uni-delta-ipv6 job to use S3 API from Ceph"

evallesp:
DNM


evallesp:
Created while investigating an error that doesn't seems to be related to this. So this is likely not necessary.

PR #753, opened 2026-05-07
nova04delta: fix preprovisioning network data for BMO

bogdando:
BMO reads preprovisioning network data from
the Secret referenced by BareMetalHost.spec.preprovisioningNetworkDataName and expects the data under the 'networkData' key.

Previously the BMH had no preprovisioningNetworkDataName (the secret was created but orphaned) and the values used 'nmstate' as the key. This did not surface because BMO never attempted to read the unreferenced secret.

- baremetalhosts.yaml: add preprovisioningNetworkDataName pointing to the co-located edpm-compute-0-preprovision-network-data Secret.
- values.yaml: rename the key from 'nmstate' to 'networkData' so the kustomize replacement produces a Secret with the key BMO reads.

Generated-by: claude-4.6-opus-high

### DIFF
--- a/dt/nova/nova04delta/edpm/baremetalhosts/baremetalhosts.yaml
+++ b/dt/nova/nova04delta/edpm/baremetalhosts/baremetalhosts.yaml
@@ -25,3 +25,4 @@ spec:
   bootMode: UEFI
   rootDeviceHints: {}
   online: true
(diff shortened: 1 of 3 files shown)


bogdando:
we are testing this again downstream

bogdando:
downstream testing confirmed this change works

abays (approved):
/lgtm
/approve

PR #755, opened 2026-05-08
Enable Barbican in multi-namespace SKMO control planes

vakwetu:
Set data.barbican.enabled to true in both central and leaf service-values so the inherited kustomize replacement maps to spec.barbican.template.enabled. The multi-namespace base leaves barbican disabled; SKMO requires Barbican in both regions per the validated scenario README.


rebtoor (approved):
/lgtm

abays (approved):
/lgtm
/approve

abays:
recheck-gate

PR #758, opened 2026-05-11
[SKMO] Add cinderBackup kustomize replacements

lkuchlan:
## Summary
- Add missing kustomize replacement rules for `cinderBackup.replicas` and `cinderBackup.customServiceConfig` in both control-plane and control-plane2
- Without these replacements, the `cinderBackup` values in `service-values.yaml` (added in #738) are never applied to the OpenStackControlPlane CR, leaving the base default of `replicas: 0`

## Test plan
- [ ] Verify `kustomize build` succeeds for both control-plane and control-plane2
- [ ] Deploy environment and confirm cinder-backup pods are running

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/va/multi-namespace-skmo/control-plane/service-values.yaml
+++ b/examples/va/multi-namespace-skmo/control-plane/service-values.yaml
@@ -17,13 +17,6 @@ data:
     customServiceConfig: |
       [DEFAULT]
       backup_driver = cinder.backup.drivers.swift.SwiftBackupDriver
-      backup_swift_auth = single_user
-      backup_swift_user = service:cinder
-      backup_swift_key = {{ .ServicePassword }}
-      backup_swift_tenant = service
-      backup_swift_auth_version = 3
-      backup_swift_project_domain = Default
-      backup_swift_user_domain = Default
     replicas: 1
   cinderVolumes:
     lvm-iscsi:
--- a/examples/va/multi-namespace-skmo/control-plane2/service-values.yaml
+++ b/examples/va/multi-namespace-skmo/control-plane2/service-values.yaml
@@ -38,13 +38,6 @@ data:
     customServiceConfig: |
       [DEFAULT]
       backup_driver = cinder.backup.drivers.swift.SwiftBackupDriver
-      backup_swift_auth = single_user
-      backup_swift_user = service:cinder
-      backup_swift_key = {{ .ServicePassword }}
-      backup_swift_tenant = service
-      backup_swift_auth_version = 3
-      backup_swift_project_domain = Default
-      backup_swift_user_domain = Default
     replicas: 1
   cinderVolumes:
     lvm-iscsi:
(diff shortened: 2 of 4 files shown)


examples/va/multi-namespace-skmo/control-plane/service-values.yaml
```diff
@@ -18,13 +17,6 @@ data:
     customServiceConfig: |
       [DEFAULT]
       backup_driver = cinder.backup.drivers.swift.SwiftBackupDriver
```
vakwetu, line 20: Just to confirm - you did intend to remove this config, right?  The default config will suffice?

examples/va/multi-namespace-skmo/control-plane2/service-values.yaml
```diff
@@ -39,13 +38,6 @@ data:
     customServiceConfig: |
       [DEFAULT]
       backup_driver = cinder.backup.drivers.swift.SwiftBackupDriver
```
vakwetu, line 40: ditto here ..

examples/va/multi-namespace-skmo/control-plane/service-values.yaml
lkuchlan: yes, exactly
the default configuration is sufficient

examples/va/multi-namespace-skmo/control-plane2/service-values.yaml
lkuchlan: this is the only required configuration for enabling the swift backup

examples/va/multi-namespace-skmo/control-plane/service-values.yaml
```diff
@@ -18,13 +18,6 @@ data:
...
-      backup_swift_auth = single_user
```
fultonj, line 21: Does the backup work with these fields removed?

The commit message should explain why they are being removed.

fultonj (dismissed):
Hi Liron,

Would you please remove this commit from the PR:

[Merge branch 'main' into skmo-cinder-backup-kustomize-fix](https://github.com/openstack-k8s-operators/architecture/pull/758/changes/e45974a0863303bc53e32a95f53ac43d35ef578f)

Merge commits clutter the history. Instead of doing the above, the PR should be rebased on main so that the PR is just the single 9704de2 commit.

https://github.com/openstack-k8s-operators/architecture/pull/758/commits


Otherwise it seems to produce the expected `oscp` yaml (aside from my question about the removed swift lines below).

```diff
         cinderBackup:
  +        customServiceConfig: |
  +          [DEFAULT]
  +          backup_driver = cinder.backup.drivers.swift.SwiftBackupDriver
           networkAttachments:
           - storage
  -        replicas: 0
  +        replicas: 1
```

examples/va/multi-namespace-skmo/control-plane/service-values.yaml
lkuchlan: Yes, tested and confirmed working. The CR is deployed and fully ready with per_user auth (Cinder's default). User tokens can reach Swift across regions via the Keystone service catalog, so the explicit single_user credentials are unnecessary.

I've updated the commit message to explain the removal

abays (approved):
/lgtm
/approve

PR #765, opened 2026-06-10
Bump pidone compute amount to 5

lmiccini:
Expand the pidone deployment topology from 3 to 5 compute nodes to support larger-scale testing scenarios.


abays (approved):
/lgtm
/approve

PR #770, opened 2026-06-18
Fix/nfv controlplane novnc and metricstorage

aharivel:
Fixes two gaps in the NFV OVS-DPDK-SRIOV controlplane configuration identified by comparing rendered manifests against authoritative RHOSO component specs:

  - **Add noVNCProxyServiceTemplate to nova cell1** (`lib/control-plane/base`): The base OpenStackControlPlane defines cell1 without a noVNCProxyServiceTemplate, meaning VNC console access to VMs on cell1 compute nodes is unavailable out of the box. The authoritative RHOSO docs specify `enabled: true` with a `ctlplane` network attachment. This affects all validated architectures inheriting from `lib/control-plane/base`.

  - **Add metricStorage kustomize replacement** (`va/nfv/ovs-dpdk-sriov`): The NFV VA already exposes `telemetry.enabled` and `ceilometer.enabled` via service-values replacements, but `metricStorage.enabled` has no replacement — it stays `false` regardless of deployer intent. This adds the missing replacement so NFV deployments can toggle Prometheus metric collection through `service-values.yaml`.

### DIFF
--- a/lib/control-plane/base/openstackcontrolplane.yaml
+++ b/lib/control-plane/base/openstackcontrolplane.yaml
@@ -204,6 +204,10 @@ spec:
             cluster: rabbitmq-cell1
           conductorServiceTemplate:
             replicas: 1
+          noVNCProxyServiceTemplate:
+            enabled: true
+            networkAttachments:
+              - ctlplane
           hasAPIAccess: true
   octavia:
     enabled: false
--- a/examples/va/nfv/ovs-dpdk-sriov-ipv6/service-values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov-ipv6/service-values.yaml
@@ -63,6 +63,8 @@ data:
     template:
       ceilometer:
         enabled: true
+      metricStorage:
+        enabled: true
   extraMounts:
     - name: v1
       region: r1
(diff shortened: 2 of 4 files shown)


mnietoji:
It looks ok to me

lib/control-plane/base/openstackcontrolplane.yaml
```diff
@@ -204,6 +204,10 @@ spec:
             cluster: rabbitmq-cell1
           conductorServiceTemplate:
             replicas: 1
+          noVNCProxyServiceTemplate:
+            enabled: true
+            networkAttachments:
+              - ctlplane
```
abays, line 210: This is effectively changing all DTs and VAs, so we would ideally need some sort of testing with a handful of associated jobs to ensure it doesn't break anything.  We probably shouldn't risk merging this blindly when we're trying to get FR6 out the door.

aharivel: @abays Ok that's understandable. I'll wait for FR6 to be out then remind you this during next merging window. Thanks!

aharivel: @abays What if I split the PR ? The novnc can wait after FR6, but maybe we can merge the second one related to metricstorage that affect only NFV ? WDYT ?

abays: Yes, that sounds reasonable to me.

aharivel:
I will split the PR in wo commit. Thanks.

PR #771, opened 2026-06-19
fix(nfv): correct enabled_filters typo in nova scheduler config

aharivel:
Misspelled as 'enabled_filtes' across all NFV VA examples (ovs-dpdk, ovs-dpdk-sriov, ovs-dpdk-sriov-ipv6, sriov). OpenStack silently ignores unknown keys, so PciPassthroughFilter and NUMATopologyFilter were never active — breaking SR-IOV and NUMA-aware scheduling.


mnietoji:
it looks ok

abays (approved):
NFV folks are good with this, so...

/lgtm
/approve

PR #773, opened 2026-06-22
feat(nfv): add metricStorage replacement 

aharivel:
The NFV OVS-DPDK-SRIOV validated architecture enables telemetry and ceilometer via service-values replacements, but metricStorage (Prometheus monitoring stack) has no replacement - it stays at the base default of enabled: false regardless of what the deployer sets in service-values.

This adds a kustomize replacement for telemetry.template.metricStorage.enabled so NFV deployments can toggle Prometheus metric collection through service-values.yaml, consistent with how telemetry.enabled and ceilometer.enabled are already handled.

Also adds the metricStorage.enabled field to example service-values for both ovs-dpdk-sriov and ovs-dpdk-sriov-ipv6 to satisfy the kustomize replacement source requirement and provide a working default.

### DIFF
--- a/examples/va/nfv/ovs-dpdk-sriov-ipv6/service-values.yaml
+++ b/examples/va/nfv/ovs-dpdk-sriov-ipv6/service-values.yaml
@@ -63,6 +63,8 @@ data:
     template:
       ceilometer:
         enabled: true
+      metricStorage:
+        enabled: true
   extraMounts:
     - name: v1
       region: r1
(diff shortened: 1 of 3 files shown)


aharivel:
Hi @mnietoji @abays 
This is the metric storage patch of the previous PR #770 
The CI is green, so you can review/approve whenever you have the time. 
Thanks !

abays (approved):
This work was previously okay'd in https://github.com/openstack-k8s-operators/architecture/pull/770, but was split out to a separate PR

/lgtm
/approve

PR #779, opened 2026-07-06
Add stage ordering, timeouts, and hook format to AGENTS.md

skovili:
The automation stage format section described the fields but lacked the conventional stage ordering, typical timeouts, a concrete stage definition example, and the pre_stage_run hook structure. Agents and contributors working with automation/vars/ files need these details to produce correct stage sequences without guessing.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,6 +82,92 @@ These files are validated against the Yamale schema at
 `.ci/automation-schema.yaml`. If you change the automation structure, update
 the schema as well.
 
+#### Standard stage ordering
+
+Stages in `automation/vars/<name>.yaml` follow this conventional order:
+
+1. `nncp-configuration` -- Node Network Configuration Policies (may include
+   pre-stage node labeling). Timeout: **5m**.
+2. `network-configuration` -- MetalLB, broader networking (wait for MetalLB
+   speaker pods).
+3. `control-plane` -- OpenStackControlPlane CR. Timeout: **60m**.
+4. `edpm-networker-nodeset` -- Networker dataplane nodeset (if applicable).
+   Timeout: **10m**.
+5. `edpm-networker-deployment` -- Networker deployment (if applicable).
+   Timeout: **40m**.
+6. `edpm-nodeset` -- Compute dataplane nodeset (wait for SetupReady).
+   Timeout: **10m**.
+7. `edpm-deployment` -- Compute dataplane deployment. Timeout: **60m**.
+
+Not all stages are required. Simple topologies may skip networker stages.
+Ceph-based topologies split dataplane into pre-ceph and post-ceph stages.
+
+#### Stage definition example
+
+Each automation vars file wraps stages under `vas: <topology-name>:` (the
+repo uses `vas:` as the top-level key for both VAs and DTs):
+
+```yaml
+vas:
+  <topology-name>:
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -82,6 +82,92 @@ These files are validated against the Yamale schema at
...
+```yaml
+vas:
+  <topology-name>:
+    stages:
+      - name: nncp-configuration
+        path: examples/dt/<topology>/control-plane/networking/nncp
+        values:
+          - name: network-values
+            src_file: values.yaml
+        build_output: nncp.yaml
```
abays, line 123: It would be good to describe what `build_output` is below.  Something like:

```
build_output -- filename for the kustomize build output artifact (used by ci-framework's kustomize_deploy role).
```

AGENTS.md
```diff
@@ -82,6 +82,98 @@ These files are validated against the Yamale schema at
...
 the schema as well.
 
+#### Standard stage ordering
+
+Stages in `automation/vars/<name>.yaml` follow this conventional order:
+
+1. `nncp-configuration` -- Node Network Configuration Policies (may include
+   pre-stage node labeling). Timeout: **5m**.
+2. `network-configuration` -- MetalLB, broader networking (wait for MetalLB
+   speaker pods).
```
abays, line 92: Let's put something like...

```
Timeout: varies (often  omitted, ~5m when set).
```

...so the agents don't guess incorrectly if they decide to add a timeout.

AGENTS.md
skovili: added

AGENTS.md
```diff
@@ -82,6 +82,98 @@ These files are validated against the Yamale schema at
...
+**`type: playbook`** -- run an Ansible playbook (used for Ceph deployment
+and similar operations that require orchestration beyond `oc` commands):
+
+```yaml
+        post_stage_run:
+          - name: "Deploy Ceph"
+            type: playbook
+            source: "../../hooks/playbooks/ceph.yml"
+            inventory: "${HOME}/ci-framework-data/artifacts/zuul_inventory.yml"
+```
```
abays: The text says "For non-core resources, api_version and namespace fields are also available" but doesn't clarify that `kind`, `resource_name`, `state`, and `definition` are the` type: cr`-specific fields, while `source` and `inventory` are `type: playbook`-specific.  The schema itself treats them all as optional on a flat `_hook` type, so an agent could incorrectly mix them.  A sentence clarifying the field groupings would help.

AGENTS.md
skovili: added a general note that the timeouts shown are the most common values — actual timeouts vary by topology, so agents should check the closest existing automation vars file rather than copying defaults blindly. is that fine?

skovili: at #102

abays: Yes, that's good

AGENTS.md
skovili: Added a field grouping clarification

abays (commented):
/lgtm

@fultonj Do you want to review as well?

fultonj (approved):
/lgtm
/approve

fultonj:
Nice work @skovili !

abays:
recheck-gate

PR #791, opened 2026-08-02
Fix invalid nexthop IP with stray unicode characters

m-blue-m:
Remove ØØ characters appended to nexthop 192.168.123.1 in subnet1 routes across bgp_dt01, bgp-l3-xl, and dz-storage NNCP values.

Assisted-By: Claude Code


fultonj (approved):
/lgtm
/approve

PR #794, opened 2026-08-03
[DNM] Add per-zone Nova cells (cell1/cell2/cell3) to dz-storage DT

gais-ameer:
Introduce Nova cells to dz-storage, making each availability zone 
its own cell with its own conductor, Galera database, and RabbitMQ cluster, 
and point each AZ's compute nodeset at its own 
cell's nova-cell[N]-compute-config secret so hosts land in the correct cell.

Signed-off-by: Gais Ameer <gameer@redhat.com>

### DIFF
--- a/examples/dt/dz-storage/control-plane/service-values.yaml
+++ b/examples/dt/dz-storage/control-plane/service-values.yaml
@@ -440,17 +440,16 @@ data:
       [DEFAULT]
       default_schedule_zone=az0
     metadataServiceTemplate:
-      enabled: true
-      # if cells per zone, set false
-      # enabled: false
+      # Set to true only when not deploying cells per zone
+      # enabled: true
+      enabled: false
     cellTemplates:
       cell0:
         cellDatabaseAccount: nova-cell0
         hasAPIAccess: true
-      # if cells per zone, uncomment topologyRef
       cell1:
-        # topologyRef:
-        #   name: azone-node-affinity
+        topologyRef:
+          name: azone-node-affinity
         cellDatabaseInstance: openstack-cell1
         cellDatabaseAccount: nova-cell1
         messagingBus:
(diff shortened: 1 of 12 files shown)


fultonj:
I'm just putting a DNM label on it for now until it's tested.

fultonj (commented):
/lgtm

examples/dt/dz-storage/control-plane/service-values.yaml
```diff
@@ -440,55 +440,52 @@ data:
...
+      enabled: false
     cellTemplates:
       cell0:
         cellDatabaseAccount: nova-cell0
         hasAPIAccess: true
-      # if cells per zone, uncomment topologyRef
       cell1:
-        # topologyRef:
-        #   name: azone-node-affinity
+        topologyRef:
```
fultonj: Could you remove the `topologyRef` for each cell service so that they inherit the default topology which is to spread across the three zones. 

That should result in this:
```
zone1                         zone2                                         zone3
cell1-db-mq-a           cell1-db-mq-b                           cell1-db-mq-c
cell2-db-mq-a           cell2-db-mq-b                           cell2-db-mq-c
cell3-db-mq-a           cell3-db-mq-b                           cell3-db-mq-c
```
Instead of this (which is what the current PR would do):
```
zone1                         zone2                                         zone3
cell1-db-mq-a           cell2-db-mq-b                           cell3-db-mq-c
cell1-db-mq-a           cell2-db-mq-b                           cell3-db-mq-c
cell1-db-mq-a           cell2-db-mq-b                           cell3-db-mq-c
```
If I temprarily loose zone1 (e.g. power outage), the API will stay up since it is spread. I want the DB for cell1 to stay up even if the compute nodes are offline.

fultonj (dismissed):
Thinking about this further, would you please remove the `topologyRef` for each rabbit and galera so that they inherit spread pods. I think we want to spread them instead of keeping them in the zone.

examples/dt/dz-storage/control-plane/service-values.yaml
gais-ameer: I have set default topology for nova and rabbitmq.
What about cinder and other services with < 3 replicas ?

fultonj: Thanks that's good. 

The distribution of other services is already correct.

fultonj (approved):
/approve
/lgtm

